PR #2, opened 2020-07-15
RDK-29161: Sky Contribution of Dobby to RDK

TeknoVenus:
Contribution from Sky of Dobby container manager code into RDK

Signed-off-by: Stephen Foulds <stephen.foulds@sky.uk>


jamcoole:
There are several compliance issues with this contribution, there are some issues with copyright headers which can be resolved at a later date and there are four code matches in blackduck protex one of which is a blocker as it is matching to GPLv3 which per policy is not allowed.

MutexTest.cpp is matching to GPLv3 and as such is blocking this PR,  the struct user_lock{} matches to GPLv3 and needs to be rewritten to be original, alternatively the test code could be removed from the contribution until the issue is resolved.

The other code matches are as follows:
AI_MD5.c/h these are already approved.  

AI_MD5_gtest.cpp is also matching but this is OK because it matches a series of magic numbers which come ultimately from RFC1321, distribution of which is unlimited. There  should be an attribution in this file and in a NOTICE file, but this can be added later.

mehrar:
Thanks @jamcoole, we've looked that issues and all seem easily resolvable. @TeknoVenus will make the required changes and re-submit

jamcoole:
@mehrar @TeknoVenus based on latest commits I will upvote the blackduck and copyright scans so you can merge as any remaining issues can be resolved post merge, we will do a full review of all remaining issues next week and perform the relevant ID work in blackduck protex, but please note more compliance related changes may be required to meet RDK policy.

mehrar:
Thanks @jamcoole 
LGTM

PR #13, opened 2020-08-07
Add networking option 'Holepuncher' to allow containers to run servers

Anttikyll:
RDK-29285


rdkPlugins/Networking/source/HolePuncher.cpp
```diff
@@ -0,0 +1,376 @@
...
+    for (int i = 0; i < len; i++)
+    {
+        // accept empty protocol which is defaulted to tcp later on
+        if (holes[i]->protocol == nullptr)
+        {
+            continue;
+        }
+
+        std::string prot = holes[i]->protocol;
+        if (strcmp(prot.c_str(), "tcp") == 0 || strcmp(prot.c_str(), "udp") == 0)
```
TeknoVenus, line 57: Would be good to convert prot to upper/lower case to allow both `TCP` and `tcp`

PR #16, opened 2020-08-11
RDK-29284 - Add localhost port forwarding for containers

Anttikyll:
The networking plugin data field 'loports' can now be used to forward ports to get access to certain ports on the host from a container.


rdkPlugins/Networking/include/PortForwarding.h
```diff
@@ -0,0 +1,106 @@
...
+#include <list>
+#include <mutex>
+#include <string>
+#include <memory>
+#include <vector>
+
+
+// -----------------------------------------------------------------------------
+/**
+ *  @namespace PortForwarding TODO: add loports
```
TeknoVenus, line 39: Remove/fix TODO

rdkPlugins/Networking/source/PortForwarding.cpp
```diff
@@ -0,0 +1,693 @@
...
+            {
+                AI_LOG_ERROR("invalid protocol value '%s' for port at index %d",
+                             ports[i]->protocol, i);
+                return false;
+            }
+        }
+        else
+        {
+            protocol = "tcp";
+        }
```
TeknoVenus, line 519: This block is duplicated from `constructHostToContainerRules`.

Perhaps refactor into `getProtocol` method?

bundle/runtime-schemas/defs-plugins.json
```diff
@@ -196,19 +196,41 @@
...
-                                    "port": {
-                                        "$ref": "defs.json#/definitions/uint16"
-                                    },
-                                    "protocol": {
-                                        "type": "string"
+                        "portForwarding": {
+                            "type": "object",
+                            "properties": {
+                                "hostToContainer": {
+                                    "type": "array",
```
Koczek9, line 203: Minor: Definition of "array of objects port+protocol" can be named better, stored and used twice in "hostToContainer" and "containerToHost". To see how it is done you could check i.e. "defs.json#/definitions/uint16"

rdkPlugins/Networking/include/PortForwarding.h
```diff
@@ -0,0 +1,106 @@
...
+
+std::vector<Netfilter::RuleSet> constructRules(const std::shared_ptr<NetworkingHelper> &helper,
+                                               const std::string &containerId,
+                                               rt_defs_plugins_networking_data_port_forwarding *portForwards,
+                                               const int ipVersion);
+
+bool constructHostToContainerRules(std::vector<Netfilter::RuleSet> &ruleSets,
+                                   const std::string &containerId,
+                                   const std::string &containerAddress,
+                                   rt_defs_plugins_networking_data_port_forwarding_host_to_container_element **ports,
```
Koczek9, line 71: I don't like using pointers to structure from automatic generated code. I think it would be better to make one part of the code to read those values and change them to "normal" c++ types like `vector<pair<int,string>>` or similar. This way if something will be changed later in the json schema only parser fragment will need to be changed, not whole logic.
Same comment applies to most of methods in this file.

rdkPlugins/Networking/source/PortForwarding.cpp
```diff
@@ -0,0 +1,693 @@
...
+                                 const std::string &ipAddress,
+                                 const std::string &portNumber,
+                                 const int ipVersion)
+{
+    char buf[256];
+
+    std::string baseRule("PREROUTING "
+                         "! -i " BRIDGE_NAME " "
+                         "-p %s "                           // protocol
+                         "-m %s "                           // protocol
```
Koczek9, line 365: -m is module, you could leave comment why protocol is used here.

rdkPlugins/Networking/source/PortForwarding.cpp
```diff
@@ -0,0 +1,693 @@
...
+                                 const int ipVersion)
+{
+    char buf[256];
+
+    std::string baseRule("FORWARD "
+                         "-d %s/%s "                // container ip address/mask
+                         "! -i " BRIDGE_NAME " "
+                         "-o " BRIDGE_NAME " "
+                         "-p %s "                   // protocol
+                         "-m %s "                   // protocol
```
Koczek9, line 432: -m is module, you could leave comment why protocol is used here.

tests/test_runner/network_tests.py
```diff
@@ -0,0 +1,128 @@
...
+        test_utils.print_single_result(output)
+
+    return test_utils.count_print_results(output_table)
+
+
+if __name__ == "__main__":
+    test_utils.parse_arguments(__file__)
+
+    # netcat is not available on RDK builds, skip test
+    if test_utils.selected_platform == test_utils.Platforms.xi_6:
```
Koczek9, line 124: This should be done inside of "execute_test" with return value:
```
if test_utils.selected_platform == test_utils.Platforms.xi_6:
    return test_utils.print_unsupported_platform(basename(__file__), test_utils.selected_platform)
```
The reason for it is that if someone wanted to include this test into some testgroup only "execute_test" will be called, and tester would not know that it is not supported in selected platform.
You can compare it to "thunder_plugin.py".

tests/test_runner/test_utils.py
```diff
@@ -76,8 +76,7 @@ def __init__(self):
                                          "--noconsole"
                                         ],
                                         universal_newlines=True,
-                                        stdout=subprocess.PIPE,
-                                        stderr=subprocess.PIPE)
+                                        stdout=subprocess.PIPE)
```
Koczek9, line 79: For DobbyDaemon all logs are printed into stderr, this is why it was added. For now it would work with only stderr=subprocess.PIPE. I left both as when it get fixed "normal" output should be in stdout so we will not need to change it. With your proposed change some tests can not shown proper messages.
Please revert this line as it was before.

bundle/runtime-schemas/defs-plugins.json
Anttikyll: I agree, this would be a good upgrade. I've just tried to do that though and it seems like there's a bug in libocispec, where if you try to add an array with defs in that way, when using the types as reference do not get the suffix "_element", causing buildtime errors. Without using defs, the suffix is added correctly. This can be seen functioning with the "loopback" element as well.

tests/test_runner/test_utils.py
Anttikyll: Good catch! That was not done on purpose, and slipped through when I reviewed my own changes. Thanks!

PR #17, opened 2020-08-11
[RDK-29360] Add --westeros-socket option

TeknoVenus:
Add the ability to pass a westeros socket path as an option when starting a container. The socket is then bind-mounted into the container with the following config:

```
        {
            "source": "/tmp/westeros-0",
            "destination": "/tmp/westeros",
            "options": [
                "bind",
                "rw",
                "nosuid",
                "nodev",
                "noexec"
            ],
            "type": "bind"
        }
```

Usage:
```
DobbyTool start --westeros-socket /tmp/westeros sleep ./sleepy
```


Anttikyll:
LGTM

PR #19, opened 2020-08-18
Update Vagrantfile to use GCC and G++ 6.4 by default

TeknoVenus:
Latest sprint branch uses GCC 6.4 - upgrade the Vagrant VM.

Set 6.4 as default instead of 5.4 included with Xenial.


Anttikyll (approved):
LGTM

PR #20, opened 2020-08-19
RDK-29415 Container kill fix

Koczek9:
This work fixes problem found in creation of "RDK-29415 Add Dobby as backend for ProcessContainers", where containers killed by SIGTERM wasn't killed properly.

### DIFF
--- a/daemon/lib/source/DobbyRunC.cpp
+++ b/daemon/lib/source/DobbyRunC.cpp
@@ -340,10 +340,33 @@ bool DobbyRunC::kill(const ContainerId& id, int signal, bool all) const
         return false;
     }
 
-    AI_LOG_FN_EXIT();
 
     // get the return code, 0 for success, 1 for failure
-    return (WEXITSTATUS(status) == EXIT_SUCCESS);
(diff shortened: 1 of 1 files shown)


daemon/lib/source/DobbyRunC.cpp
```diff
@@ -340,10 +340,33 @@ bool DobbyRunC::kill(const ContainerId& id, int signal, bool all) const
...
     // get the return code, 0 for success, 1 for failure
-    return (WEXITSTATUS(status) == EXIT_SUCCESS);
+    bool returnValue = (WEXITSTATUS(status) == EXIT_SUCCESS);
+
+    // Fix problem where SIGTERM was masked and containers never exited
+    if(signal == SIGTERM)
+    {
+        int retryCounter = 10;
+
+        while (state(id) != ContainerStatus::Unknown && retryCounter > 0)
```
TeknoVenus, line 352: Would it also make sense to check if `state(id != ContainerStattus::Stopped` - since I would assume if the state is Stopped the container was also killed successfully?

Koczek9: Fixed.

TeknoVenus (approved):
LGTM

PR #21, opened 2020-08-20
RDK-29415 - Add Dobby as backend for ProcessContainers

Koczek9:
Removed dependency of jsoncpp.

### DIFF
--- a/AppInfrastructure/Public/Dobby/IDobbyProxy.h
+++ b/AppInfrastructure/Public/Dobby/IDobbyProxy.h
@@ -26,10 +26,6 @@
 
 #include <Common/Notifier.h>
 
-#if defined(RDK)
-#  include <json/json.h>
-#endif
-
 #include <cstdint>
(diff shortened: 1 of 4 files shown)


AppInfrastructure/Public/Dobby/IDobbyProxy.h
```diff
@@ -28,8 +28,6 @@
 
 #if defined(RDK)
 #  include <json/json.h>
```
TeknoVenus, line 30: Why is this still here if we've removed the jsoncpp as a dependency?

TeknoVenus (changes requested):

AppInfrastructure/Public/Dobby/IDobbyProxy.h
Koczek9: Fixed

TeknoVenus (approved):
LGTM

PR #22, opened 2020-08-20
Networking plugin improvements

Anttikyll:
- Fix multiple iptables/ip6tables fork/exec executions for significant time reductions in networking setup
- Fix bug with dnsmasq enabling iptables rules not being removed on the xi6
- Fix bug with portForwards options causing SEGFAULTs in some situations

### DIFF
--- a/rdkPlugins/Networking/source/Netfilter.cpp
+++ b/rdkPlugins/Networking/source/Netfilter.cpp
@@ -421,22 +421,110 @@ Netfilter::RuleSet Netfilter::getRuleSet(const int ipVersion) const
 
 // -----------------------------------------------------------------------------
 /**
- *  @brief Uses the iptables-restore tool to apply the given rules
+ *  @brief Trims duplicates from mRuleSets based on the operation.
+ *
+ *  Rules with 'Delete' operation will be removed from mRuleSets if the rule is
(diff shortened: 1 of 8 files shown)


rdkPlugins/Networking/source/Netfilter.cpp
```diff
@@ -679,257 +778,159 @@ bool Netfilter::ruleInList(const std::string &rule,
...
+ *  These rules will be prefixed with "-D" to signal deleting from iptables.
  *
  *  @param[in]  ruleSet         The ruleset to apply.
  *  @param[in]  ipVersion       iptables version to use.
  *
- *  @return true on success, false on failure.
+ *  @return always returns true.
  */
-bool Netfilter::deleteRules(const RuleSet &ruleSet, const int ipVersion)
+bool Netfilter::deleteRules(RuleSet &ruleSet, const int ipVersion)
```
Koczek9, line 866: This name is very confusing as "deleteRules" as this functon is adding more elements, not removing them is more like "addDeleteRules". Also those all "deleteRules", "appendRules", "insertRules" looks like they have so similar logic that if written correctly they could be one function, but I think that it is not worth changing as you wrote this one already. So in my opinion just change naming to "addDeleteRules" and other probably too, to be consistent.

PR #23, opened 2020-08-21
Various usability improvements

Anttikyll:
- Removed debugging log bloat caused by `DobbyWorkQueue`
- Allow underscores in container names
- Fixed try/catch misuse in DobbyTool
- Improved error handling for `DobbyTool start` with dodgy paths

### DIFF
--- a/daemon/lib/source/DobbyWorkQueue.cpp
+++ b/daemon/lib/source/DobbyWorkQueue.cpp
@@ -111,8 +111,6 @@ bool DobbyWorkQueue::runFor(const std::chrono::milliseconds &msecs)
  */
 bool DobbyWorkQueue::runUntil(const std::chrono::steady_clock::time_point &deadline)
 {
-    AI_LOG_FN_ENTRY();
-
     // wait for the next work item added or mTerminateWorkQueue is true
     const auto predicate = [&]()
     {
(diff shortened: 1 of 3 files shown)


daemon/lib/source/DobbyWorkQueue.cpp
```diff
@@ -111,8 +111,6 @@ bool DobbyWorkQueue::runFor(const std::chrono::milliseconds &msecs)
  */
 bool DobbyWorkQueue::runUntil(const std::chrono::steady_clock::time_point &deadline)
 {
-    AI_LOG_FN_ENTRY();
```
Koczek9, line 114: If you are commenting out `AI_LOG_DEBUG("waiting for next work item");` this should be also commented out. Preferably with comment stating "Commented out, because this method is called so often that it shadows all other debugs" or something similar (so your intention is clear, and doesn't get "corrected" in future).

TeknoVenus (approved):

daemon/lib/source/DobbyWorkQueue.cpp
Anttikyll: There was no reason to leave the entry/exits there. I was umming and ahing on whether or not to get rid of the debug line as well, but ultimately decided to keep it to follow the trend that's been going on. I'm happy to get rid of it as well, since if someone wants to see debugs on every print, they can just write a debug print.

Koczek9:
LGTM

TeknoVenus (approved):
LGTM

PR #24, opened 2020-08-25
Add link time optimisation build option

Anttikyll:
Using LTO drastically reduces the size of installed files

Enable or disable by using cmake build option ENABLE_LTO


TeknoVenus:
LTO can cause issues with debugging symbols with GCC <8 (http://hubicka.blogspot.com/2018/06/gcc-8-link-time-and-interprocedural.html) so ENABLE_LTO should be off by default for Debug builds.

Of course, this then leads into the fact Dobby currently doesn't compile under a Release build which needs fixing at some point

Anttikyll:
Great find! I'll turn it off by default and add a comment.

TeknoVenus (approved):
LGTM

PR #26, opened 2020-08-26
Add syshooks and dobby specs to LEGACY_COMPONENTS

Anttikyll:
Allows removal of ctemplate dependency for builds where dobby
specs are not needed.

With LEGACY_PLUGIN=OFF, Dobby syshooks, legacy plugins and dobby
specs no longer function.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,11 @@ endif()
 # Add defines for DBUS path
 add_definitions( -DDBUS_SYSTEM_ADDRESS="unix:path=/var/run/dbus/system_bus_socket" )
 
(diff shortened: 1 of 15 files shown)


Anttikyll:
PR #25 brings in more legacy plugins and the gpu syshook is not yet converted to an RDK plugin, which would make merging this PR a bit too early.

Leaving as draft until the gpu syshook has been brought in as an rdk plugin so we can properly strip these legacy components from builds without `LEGACY_PLUGIN=ON`.

We'll also need to strip down the `findRdkPlugins`/`mRdkPluginsInDevelopment` bits of code in `DobbyConfig` afterwards.

PR #27, opened 2020-08-26
RDKCMF-8664: fix for configs with zero rdk plugins

sverkoye:
* without this patch I would get this error when using a config
  with zero plugins:
  "Could not run preCreation hook as plugin manager is null"


TeknoVenus (approved):
Thanks for the PR. LGTM.

PR #28, opened 2020-08-27
make a backup of the original config.json

goruklu:
We need to be able to check whether there is an update to the original OCI config. having a backup allows us to do checksum comparisons.
Signed-off-by: Gurdal Oruklu <gurdal_oruklu@comcast.com>

### DIFF
--- a/bundle/lib/source/DobbyConfig.cpp
+++ b/bundle/lib/source/DobbyConfig.cpp
@@ -25,7 +25,7 @@
 
 #include <glob.h>
 #include <sys/stat.h>
-
+#include <fstream>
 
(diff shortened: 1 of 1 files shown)


mehrar (approved):
LGTM

Anttikyll:
The `config-dobby.json` file written to the bundle path on container launch is the original backup of the config file.

It's written with `writeConfigJsonImpl()` on line 658, just before calling `updateBundleConfig()`, which now writes the backup as well as `config-orig.json`.

IMO this adds an unnecessary write and config file, because the backup config file is already there.

PR #30, opened 2020-09-02
Change original config backup write to avoid writing twice

Anttikyll:
This moves the creation of config-orig.json to config-dobby.json
so that the written config-dobby.json is a direct copy of the
original configuration file.

See https://github.com/rdkcentral/Dobby/pull/28


TeknoVenus:
LGTM

goruklu (approved):

PR #31, opened 2020-09-02
Fix CMake libocispec command sequence

sverkoye:
* commands in execute_process() are executed in parallel
  so making the target dirs and copying files in them
  must be sequential

### DIFF
--- a/cmake/libocispec.cmake
+++ b/cmake/libocispec.cmake
@@ -70,6 +70,10 @@ execute_process(
     WORKING_DIRECTORY ${LIBOCISPEC_DIR}
     COMMAND mkdir -p ./schemas/rt
     COMMAND mkdir -p ./schemas/img
+)
+
+execute_process(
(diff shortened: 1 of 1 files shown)


TeknoVenus:
Nice spot! Thanks for the fix - LGTM

Please sign the CLA and/or fix the commit author so it can be merged.

sverkoye:
Hmm something seems wrong with this CLA thing. When I click the link it says I signed it (which I did weeks ago). But in this pull request it also says: "Stefan Verkoyen seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA". Well, my github handle is "sverkoye" which is indicate the pull request...strange. It seems to use my full name instead.

TeknoVenus:
Re-checked the CLA and it's worked this time - all checks are passing now.

TeknoVenus (approved):

PR #33, opened 2020-09-03
Replace GpuMemoryHook with GPU rdkPlugin

Anttikyll:
This is the last one of the original legacy syshook to be converted to an rdk plugin, so we can also remove
old bits of code that allow using syshooks with rdk plugins when launching from bundles.


rdkPlugins/GPU/source/GpuPlugin.cpp
```diff
@@ -0,0 +1,308 @@
...
+#include <fcntl.h>
+#include <unistd.h>
+#include <mntent.h>
+
+REGISTER_RDK_PLUGIN(GpuPlugin);
+
+GpuPlugin::GpuPlugin(std::shared_ptr<rt_dobby_schema> &containerConfig,
+                     const std::shared_ptr<DobbyRdkPluginUtils> &utils,
+                     const std::string &rootfsPath)
+    : mName("GpuPlugin"),
```
TeknoVenus, line 35: This should just be `Gpu` otherwise the rdkplugins config will need to be:
```
gpuPlugin: {
   ...
}
```
which is different to the schema

rdkPlugins/GPU/source/GpuPlugin.cpp
```diff
@@ -0,0 +1,308 @@
...
+ */
+bool GpuPlugin::postHalt()
+{
+    AI_LOG_FN_ENTRY();
+
+    // sanity check we have a gpu cgroup dir
+    if (mCgroupDirPath.empty())
+    {
+        AI_LOG_ERROR_EXIT("no gpu cgroup directory found");
+        return false;
```
TeknoVenus, line 105: I don't think this should return false - if the directory isn't found then there's just nothing to do, not a fatal error.

Otherwise causes postHalt hook failure:
```
0000007853.871211 NFO: < M:DobbyRdkPluginManager.cpp F:runPlugins L:592 > Running gpu plugin
0000007853.871262 ERR: < M:GpuPlugin.cpp F:postHalt L:104 > no gpu cgroup directory found
0000007853.871283 ERR: < M:DobbyRdkPluginManager.cpp F:runPlugins L:600 > Required plugin gpu postHalt hook has failed
0000007853.871302 ERR: < M:DobbyManager.cpp F:onPostHaltHook L:1907 > Failure in postHalt hook
```

rdkPlugins/GPU/source/GpuPlugin.cpp
```diff
@@ -0,0 +1,308 @@
...
+    AI_LOG_FN_ENTRY();
+
+    // try and do the bind mount
+    if (mount(source.c_str(), target.c_str(), nullptr, MS_BIND, nullptr) < 0)
+    {
+        AI_LOG_SYS_ERROR(errno, "failed to bind mount '%s' to '%s'",
+                         source.c_str(), target.c_str());
+        return false;
+    }
+    else
```
TeknoVenus, line 234: Can simplify by removing else statement as we return false inside the above if statement

rdkPlugins/GPU/source/GpuPlugin.cpp
```diff
@@ -0,0 +1,308 @@
...
+    {
+        AI_LOG_DEBUG("bind mounted '%s' to '%s'", source.c_str(),
+                     target.c_str());
+        return true;
+    }
+
+    AI_LOG_FN_EXIT();
```
TeknoVenus, line 241: Unreachable

TeknoVenus (dismissed):
LGTM

pluginLauncher/lib/source/DobbyRdkPluginUtils.cpp
```diff
@@ -59,19 +59,27 @@ DobbyRdkPluginUtils::~DobbyRdkPluginUtils()
...
 {
     if (stdin.empty())
     {
+        AI_LOG_ERROR_EXIT("container stdin empty - couldn't get pid");
         return 0;
     }
 
     // get pid from hook's stdin json '"pid":xxxxx'
     std::size_t pidPosition = stdin.find("\"pid\":") + 6;
+    if (pidPosition == std::string::npos)
```
TeknoVenus, line 77: If `find()` fails to find pid, we still add 6. `npos` is `-1`, so if the PID isn't found `pidPosition` would equal 5 and this error check would never be called.

The `+ 6` should be moved after this check

rdkPlugins/GPU/source/GpuPlugin.cpp
```diff
@@ -62,13 +61,15 @@ unsigned GpuPlugin::hookHints() const
...
 {
+    const std::string cgroupDirPath = getGpuCgroupMountPoint();
+
     // if we don't have a GPU cgroup controller then nothing to do
-    if (mCgroupDirPath.empty())
+    if (cgroupDirPath.empty())
     {
         AI_LOG_ERROR_EXIT("missing gpu cgroup directory");
-        return true;
+        return false;
```
TeknoVenus, line 72: Why has this flipped?

Anttikyll: If the cgoup directory isn't present, we should return with an error.

We could return true as a sign of "oh, you don't have the cgroup so no worries!", but that's a bit misleading IMO.

An upcoming commit will change the one in `postHalt()` as well.

Anttikyll:
Tested on Xi1 sky build and confirmed working.

TeknoVenus (approved):

PR #34, opened 2020-09-04
Multiple plugins fix

Koczek9:
Now there is possibility to have more than one plugin with the same name in the plugin directory. Fixed 2 separate problems where this caused Segmentation Fault.
1. Logger item was tried to be deleted from list - which called destructor, but this operation was after library containing this destructor was already closed.
2. Plugins were loaded in random order - which meant sometimes older plugin was chosen, which resulted in calls to unexciting files.


TeknoVenus (approved):
LGTM

PR #36, opened 2020-09-04
Cpp check fixes

Koczek9:
Multiple cppcheck issues fixed.


plugins/EthanLog/source/EthanLogClient.cpp
```diff
@@ -312,8 +312,8 @@ bool EthanLogClient::shouldDrop()
 
         char messageBuf[128];
         snprintf(messageBuf, sizeof(messageBuf),
-                 "MESSAGE=Dropped %u log messages in last %lld seconds (most "
-                 "recently, %lld seconds ago) due to excessive rate",
+                 "MESSAGE=Dropped %u log messages in last %ld seconds (most "
+                 "recently, %ld seconds ago) due to excessive rate",
```
anttikyl, line 316: I believe we should switch to `PRId64` instead of `%lld` here. The `%lld` -> `%ld` change system dependent, and on 32 bit systems we'd get this same warning again but the other way around.

Refer to:

[https://linux.m2osw.com/c-printf3-int-long-and-long-long-works-every-time](https://linux.m2osw.com/c-printf3-int-long-and-long-long-works-every-time)

client/lib/source/Upstart.cpp
```diff
@@ -96,7 +96,7 @@ bool Upstart::invokeMethod(const std::string& method,
             (std::chrono::steady_clock::now() - startTime);
     if (timeTaken > std::chrono::seconds(2))
     {
-        AI_LOG_WARN("upstart request took a rather long time (%llums)",
+        AI_LOG_WARN("upstart request took a rather long time (%ldms)",
```
anttikyl, line 99: I believe we should switch to PRId64 instead of %lld here. The %lld -> %ld change system dependent, and on 32 bit systems we'd get this same warning again but the other way around.

Refer to:

https://linux.m2osw.com/c-printf3-int-long-and-long-long-works-every-time

anttikyl (changes requested):

plugins/EthanLog/source/EthanLogClient.cpp
Koczek9: Fixed

client/lib/source/Upstart.cpp
Koczek9: Fixed

anttikyl (commented):
LGTM

anttikyl (approved):

PR #37, opened 2020-09-08
BD-821-BLDK-734: Updates In Licences Header

ravish-cts:
Updated License header (removal)  RDK Management copyrights which should be changed to Sky UK and the second one then removed


TeknoVenus (approved):
LGTM

PR #42, opened 2020-09-11
RDK-29415 Removed more dependency from static DobbyClient library.

Koczek9:
Now finished static library should not require linking of AppInfraLogging, AppInfraCommon, and IpcService.


TeknoVenus:
CMake files were updated as part of: https://github.com/rdkcentral/Dobby/pull/108 so this PR is no longer relevant

PR #53, opened 2020-09-30
RDK-29823 - Port Sky's MulticastForwarder plugin into RDK Networking plugin

anttikyl:
### Description

This adds a `multicastForwarding` field in the data of the `Networking` rdkPlugin. If filled correctly, it will add the required smcroute, ebtables and iptables rules to allow forwarding multicast traffic from a specific address to the container.

#### Prerequisites

You will need to install `ebtables` and `smcroute` on the VM, since they're not installed by default. They're both available with `apt-get install`.

You will also need to start the `smcroute` daemon with `smcroute -d`.

#### To run test:

1. Fill in the `rdkPlugins.networking.data.multicastForwarding` section of a container's config (see rdkPlugins/Networking/README.md for instructions).
2.  Launch container (for example sleepy with enough sleep time to check test results)
3. Check iptables mangle rule has been written correctly with `iptables -S -t mangle | grep [containername]`

The output should look something like:
```
-A PREROUTING -d 225.1.2.3/32 ! -i dobby0 -p udp -m udp --dport 1900 -m comment --comment mcf -j TTL --ttl-inc 1
```

4. Check iptables filter rule has been written correctly with `iptables -S -t filter | grep [containername]`

The output should look something like:
```
-A FORWARD -d 225.1.2.3/32 ! -i dobby0 -o dobby0 -p udp -m udp --dport 1900 -m comment --comment mcf -j ACCEPT
```
5. Check ebtables with `ebtables --list`

The output should look something like:
```
Bridge table: filter

Bridge chain: INPUT, entries: 0, policy: ACCEPT

Bridge chain: FORWARD, entries: 0, policy: ACCEPT

Bridge chain: OUTPUT, entries: 1, policy: ACCEPT
-p IPv4 -o veth0 --ip-dst 225.1.2.3 -j ACCEPT 
```

6. Check smcroute for added routes with `smcroutectl show`

The output should look something like:
```
ROUTE (S,G)                        INBOUND             PACKETS      BYTES  OUTBOUND                        
(*, 225.1.2.3)                     enp0s3                    0          0  dobby0
(*, 225.1.2.3)                     enp0s8                    0          0 
(192.168.178.36, 239.255.255.250)  enp0s8                    3        549 
(192.168.178.1, 239.255.255.250)   enp0s8                    4        604 
```

#### Test notes

This will also need to be tested properly in use with Sky Netflix for potential regression outside of this PR's scope.

### Type of Change

- [x] The base Bitbake recipe (`meta-rdk-ext/recipes-containers/dobby/dobby.bb`) must be modified to support the changes in this PR (beyond updating `SRC_REV`)

NOTE: MulticastForwarder will add a dependency for `ebtables` and `smcroutectl` on RDK. The build will run fine, but will return error messages if either is not available on box when `multicastForwarding` is present in the plugin data, so technically bitbake recipe changes are not necessary, unless if this feature will be used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,6 +246,7 @@ option(PLUGIN_TESTPLUGIN "Include TestPlugin plugin" OFF)
 option(PLUGIN_GPU "Include GPU plugin" OFF)
 option(PLUGIN_LOCALTIME "Include LocalTime plugin" OFF)
 option(PLUGIN_RTSCHEDULING "Include RtScheduling plugin" OFF)
+option(PLUGIN_HTTPPROXY "Include HttpProxy plugin" OFF)
 
 if(PLUGIN_TESTPLUGIN)
     add_subdirectory(rdkPlugins/TestPlugin)
@@ -279,6 +280,9 @@ if(PLUGIN_RTSCHEDULING)
     add_subdirectory(rdkPlugins/RtScheduling)
 endif()
 
+if(PLUGIN_HTTPPROXY)
+    add_subdirectory(rdkPlugins/HttpProxy)
+endif()
 
 # Export targets in Dobby package
 # ---------------------------------------------------------
--- a/bundle/runtime-schemas/defs-plugins.json
+++ b/bundle/runtime-schemas/defs-plugins.json
@@ -350,6 +350,46 @@
                 }
             }
         },
+        "HttpProxy": {
+            "required": [
+                "required",
+                "data"
+            ],
+            "properties": {
+                "required": {
+                    "type": "boolean"
+                },
+                "data": {
+                    "type": "object",
+                    "properties": {
+                        "proxy": {
+                            "type": "object",
+                            "required": [
+                                "host",
+                                "port"
+                            ],
+                            "properties": {
+                                "host": {
+                                    "type": "string"
+                                },
+                                "port": {
+                                    "$ref": "defs.json#/definitions/uint16"
+                                }
+                            }
+                        },
+                        "ignoreProxy": {
+                            "$ref": "defs.json#/definitions/ArrayOfStrings"
+                        },
+                        "ignoreProxyOnBridge": {
+                            "type": "boolean"
+                        },
+                        "proxyRootCACert": {
+                            "type": "string"
+                        }
+                    }
+                }
+            }
+        },
         "TestRdkPlugin": {
             "type": "object",
             "required": [
@@ -425,6 +465,9 @@
                 },
                 "rtscheduling": {
                     "$ref": "#/definitions/RtScheduling"
+                },
+                "httpproxy": {
+                    "$ref": "#/definitions/HttpProxy"
                 },
                 "testrdkplugin": {
                     "$ref": "#/definitions/TestRdkPlugin"
--- a/bundle/runtime-schemas/defs-plugins.json
+++ b/bundle/runtime-schemas/defs-plugins.json
@@ -234,6 +234,24 @@
                                     }
                                 }
                             }
+                        },
+                        "multicastForwarding": {
+                            "type": "array",
+                            "items": {
+                                "type": "object",
+                                "required": [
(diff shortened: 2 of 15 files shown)


TeknoVenus (commented):
Code LGTM.

Following your test instructions, things seem to work as you described. I did notice this in the container log though - is this expected?

```
0000019886.503666 NFO: < M:Netlink.cpp F:addRoute L:1522 > adding route 'inet 127.0.0.0/8 table main via 0.0.0.0 dev 1 '
0000019886.503824 NFO: < M:Netlink.cpp F:addRoute L:1644 > adding route 'inet6 2080:d0bb:1e::6440:b02 table main via dev 4 '
Daemon error: Invalid output interface
Daemon error: Invalid output interface
0000019886.646879 NFO: < M:DobbyRdkPluginManager.cpp F:runPlugins L:634 > Plugin networking has createRuntime hook run successfully
```

Didn't look like it affected anything though.

anttikyl:
There's an issue with `smcroute` where the daemon only refreshes the interface list on startup.

Because Dobby only adds the `dobby0` bridge interface when a container with networking plugin type none or nat is first launched, `smcrouted`'s interface list does not contain it, and adding a multicast route with `smcroutectl` fails.

The following patch can be used for `smcroute` **version 2.4.4** to fix the issue so that the interface list is updated if it's missing and a route addition is attempted for `dobby0`:

```
From aad68441389eab0a7df3f0d78a3765b88388cf58 Mon Sep 17 00:00:00 2001
From: Antti Kyllonen <antti.kyllonen@consult.red>
Date: Fri, 2 Oct 2020 13:54:49 +0000
Subject: [PATCH] Configure missing dobby0 on route add

---
 src/ifvc.h |  1 +
 src/msg.c  | 13 +++++++++++++
 2 files changed, 14 insertions(+)

diff --git a/src/ifvc.h b/src/ifvc.h
index 74ba782..cfa5c44 100644
--- a/src/ifvc.h
+++ b/src/ifvc.h
@@ -27,6 +27,7 @@ struct ifmatch {
 void          iface_init              (void);
 void          iface_refresh           (void *arg);
 void          iface_exit              (void);
+int           iface_update            (int refresh);
 
 struct iface *iface_iterator          (int first);
 
diff --git a/src/msg.c b/src/msg.c
index 1dcb459..d4bc689 100644
--- a/src/msg.c
+++ b/src/msg.c
@@ -176,6 +176,19 @@ static int do_mroute4(struct ipc_msg *msg)
 			while (pos < msg->count) {
 				char *ifname_out = msg->argv[pos++];
 
+				// smcrouted only updates the list of interfaces on startup. However, the Dobby Networking
+				// plugin doesn't create its bridge device until the first container that needs it is
+				// launched. This means that when "smcroutectl add" is called, the Dobby bridge device is
+				// unlikely to be known to smcrouted.
+				//
+				// To fix this, we update the interface list here if dobby0 is needed and missing. Then we
+				// add the virtual interface in smcrouted, which then allows routes to be added for dobby0.
+				if (strcmp(ifname_out, "dobby0") == 0 && iface_find_by_name("dobby0") == NULL)
+				{
+					iface_update(0);
+					mroute_add_vif("dobby0", 0, DEFAULT_THRESHOLD);
+				}
+
 				iface_match_init(&state_out);
 				while ((vif = iface_match_vif_by_name(ifname_out, &state_out, NULL)) >= 0) {
 					if (vif == mroute.inbound) {
-- 
2.7.4
```

Leaving PR as draft for now, pending decisions.

anttikyl:
There's also a bug on the VM, where using `smcroute` as a systemd service fails with the following error:

```
Oct 02 10:36:26 dobby-vagrant smcroute[5308]: Failed binding IPC socket, client disabled: Read-only file system
```

Running smcrouted in another window with `sudo /usr/local/sbin/smcrouted -n -l debug` works though.

TeknoVenus (approved):
LGTM

PR #54, opened 2020-10-05
RDK-29862 - Add rdkPlugins field to Dobby specs

anttikyl:
### Description

Adds the ability to add 1-to-1 copy of the `rdkPlugins` field into Legacy Dobby specs.

### Test Procedure

Use `-DLEGACY_COMPONENTS=ON` flag.

Use `DobbyBundleGenerator` to generate bundles from specs. Test adding different RDK plugins into the  `rdkPlugins` field to see if they get merged in properly.

It's expected that any plugin data given in the `rdkPlugins` field is merged **over** the data that is processed from the other fields of the spec. I.e. if you have the following Dobby spec:

```
{
    ...,
    "network": "nat",
    "rdkPlugins": {
        "networking": {
            "required": true,
            "data": {
                "type": "open",
                "ipv6": true
            }
        }
    }
}
```

The output bundle's OCI config `rdkPlugins.network` field should be:

```
        "networking": {
            "required": true,
            "data": {
                "ipv4": true,
                "ipv6": true,
                "type": "open"
            }
        }
```
Note that the `required` field is also true since it was overwritten by the spec's `rdkPlugins` field.

### Type of Change

### DIFF
--- a/bundle/lib/source/DobbySpecConfig.cpp
+++ b/bundle/lib/source/DobbySpecConfig.cpp
@@ -539,7 +539,7 @@ bool DobbySpecConfig::parseSpec(ctemplate::TemplateDictionary* dictionary,
         ProcessorMap::const_iterator fn = processors.find(it.name());
         if (fn != processors.end())
         {
-            AI_LOG_INFO("Processing %s", fn->first.c_str());
+            AI_LOG_DEBUG("Processing %s", fn->first.c_str());
             // the following is super ugly, but allows us to call the function
             // pointers as methods within our own class
             const ProcessorFunc& method = fn->second.second;
--- a/bundle/lib/source/DobbySpecConfig.cpp
+++ b/bundle/lib/source/DobbySpecConfig.cpp
@@ -135,7 +135,7 @@ static const ctemplate::StaticTemplateString ENABLE_RDK_PLUGINS =
     STS_INIT(ENABLE_RDK_PLUGINS, "ENABLE_RDK_PLUGINS");
 static const ctemplate::StaticTemplateString RDK_PLUGIN_SECTION =
     STS_INIT(RDK_PLUGIN_SECTION, "RDK_PLUGIN_SECTION");
-    static const ctemplate::StaticTemplateString RDK_PLUGIN_NAME =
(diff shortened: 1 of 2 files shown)


bundle/lib/source/DobbySpecConfig.cpp
```diff
@@ -2045,12 +1976,8 @@ bool DobbySpecConfig::processMounts(const Json::Value& value,
...
     if (numLoopMounts > 0)
     {
-        ctemplate::TemplateDictionary* subDict;
-        enableRdkPlugin(subDict, RDK_STORAGE_PLUGIN_NAME, false);
-
-        std::string pluginData = createRdkPluginDataString(rdkPluginData);
-        subDict->SetValue(RDK_PLUGIN_DATA, pluginData);
-        addRdkPlugin(RDK_STORAGE_PLUGIN_NAME, false, rdkPluginData);
+        mRdkPluginsJson[RDK_STORAGE_PLUGIN_NAME]["data"] = rdkPluginData;
+        mRdkPluginsJson[RDK_NETWORK_PLUGIN_NAME]["required"] = false;
```
TeknoVenus, line 1980: This should be `RDK_STORAGE_PLUGIN_NAME`

bundle/lib/source/DobbySpecConfig.cpp
```diff
@@ -2580,3 +2507,108 @@ bool DobbySpecConfig::processCapabilities(const Json::Value& value,
...
+ *
+ *  @return true if correctly processed the value, otherwise false.
+ */
+void DobbySpecConfig::insertIntoRdkPluginJson(const std::string& pluginName,
+                                              const Json::Value& pluginData)
+{
+    Json::Value& existingData = mRdkPluginsJson[pluginName]["data"];
+
+    // iterate through all data members in the RDK plugin's data field
+    for (auto dataMember : pluginData.getMemberNames())
```
TeknoVenus, line 2532: Same with all these for loops - any reason not to use `const auto&` here since you're not modifying the contents of dataMember and I see no real reason to make copies?

TeknoVenus (changes requested):

TeknoVenus (approved):
LGTM!

PR #55, opened 2020-10-06
Fixed bug causing plugin launcher test to freeze.

Koczek9:
### Description
Test was waiting for pid of container, which normally is provided in stdin by crun. In this case crun was not providing it which caused infinite wait. Now test provides this pid field itself.

### Test Procedure
Go to ~/srcDobby/tests/test_runner/
run "python3 plugin_launcher.py -p1|

Without this patch this command will freeze forever. With this patch there should be "All 1 succeed" message.

### Type of Change
- Bug fix

- none


tests/test_runner/plugin_launcher.py
```diff
@@ -27,6 +28,7 @@
                     "Runs plugins launcher and check if it runned properly"),
 )
 
+crun_input = '{"pid":12345}'
```
TeknoVenus, line 31: We know from the OCI spec that crun will provide the container state over stdin in:

> The state of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.

and the state is defined as

> When serialized in JSON, the format MUST adhere to the following pattern:
> ```
> {
>    "ociVersion": "1.0.2",
>    "id": "oci-container1",
>    "status": "running",
>    "pid": 4422,
>    "bundle": "/containers/redis",
>    "annotations": {
>        "myKey": "myValue"
>    }
> }
> ```

Would probably be worth using this whole json block rather than just the PID for a more accurate test

Koczek9: Added full JSON instead short one.

anttikyl (approved):
LGTM

Tested and output is as expected.

PR #57, opened 2020-10-09
Check compatiblity when removing multicast forwarder rules in postHalt

TeknoVenus:
### Description
Add a check in MulticastForwarder to check compatibility when removing rules. Previously only checked compatibility when adding the rules.

### Test Procedure
Start/stop a container when `smcroute` isn't installed. Without the patch, you will see the following error in DobbyDaemon when stopping the container:

```
sh: 1: /usr/sbin/smcroutectl: not found
sh: 1: /usr/sbin/smcroutectl: not found
Sorry, rule does not exist.
```

With the patch, you will see a nicer error:

```
Multicast forwarding not supported - smcroutectl not found in PATH (2 - No such file or directory)
```

### Type of Change


TeknoVenus:
@Koczek9 Please can you review - would like to merge alongside https://github.com/rdkcentral/Dobby/pull/63

Koczek9 (approved):
LGTM

PR #59, opened 2020-10-12
Fix invalid data for Local Time rdk plugin

TeknoVenus:
### Description
Fix issue where the data section of the `localtime` plugin was invalid, preventing container startup on Xi1

### Test Procedure
Attempt to start containers on Xi1. Containers should start without error, and the localtime plugin in the config.json should look as follows:

```json
"localtime":{
      "required":false,
      "data":{}
   }
```

### Type of Change

### DIFF
--- a/bundle/lib/source/DobbySpecConfig.cpp
+++ b/bundle/lib/source/DobbySpecConfig.cpp
@@ -618,7 +618,8 @@ bool DobbySpecConfig::parseSpec(ctemplate::TemplateDictionary* dictionary,
     // step 6.5 - enable localtime rdk plugin by default if on Xi1. The
     // localtime plugin takes no input params, so we simply enable the
     // rdkPlugin rather than processing it via a processing function.
-    mRdkPluginsJson[RDK_LOCALTIME_PLUGIN_NAME]["data"] = "{}";
+    Json::Value rdkPluginData = Json::objectValue;
+    mRdkPluginsJson[RDK_LOCALTIME_PLUGIN_NAME]["data"] = rdkPluginData;
     mRdkPluginsJson[RDK_LOCALTIME_PLUGIN_NAME]["required"] = false;
 #endif
 


bundle/lib/source/DobbySpecConfig.cpp
```diff
@@ -618,7 +618,8 @@ bool DobbySpecConfig::parseSpec(ctemplate::TemplateDictionary* dictionary,
     // step 6.5 - enable localtime rdk plugin by default if on Xi1. The
     // localtime plugin takes no input params, so we simply enable the
     // rdkPlugin rather than processing it via a processing function.
-    mRdkPluginsJson[RDK_LOCALTIME_PLUGIN_NAME]["data"] = "{}";
+    Json::Value rdkPluginData = Json::objectValue;
+    mRdkPluginsJson[RDK_LOCALTIME_PLUGIN_NAME]["data"] = rdkPluginData;
```
Koczek9, line 622: LGTM, also I think that just using Json::objectValue should be enough (compare jsoncpp/include/json/value.h line 312), but your solution will always work without doubts.

Koczek9 (commented):
To be honest I get 
```
"localtime":{
      "required":false,
      "data":{}
   }
```
result with my unchanged XiOne. It is two week old build, so was this some kind of regression?

bundle/lib/source/DobbySpecConfig.cpp
TeknoVenus: > To be honest I get
> 
> ```
> "localtime":{
>       "required":false,
>       "data":{}
>    }
> ```
> 
> result with my unchanged XiOne. It is two week old build, so was this some kind of regression?

Yes, regression was introduced in https://github.com/rdkcentral/Dobby/pull/54

TeknoVenus:
> To be honest I get
> 
> ```
> "localtime":{
>       "required":false,
>       "data":{}
>    }
> ```
> 
> result with my unchanged XiOne. It is two week old build, so was this some kind of regression?

Yes, regression was introduced in https://github.com/rdkcentral/Dobby/pull/54

Koczek9 (dismissed):
LGTM

Koczek9 (approved):

PR #60, opened 2020-10-22
Build Dobby in Release mode with libocispec as a shared library

TeknoVenus:
### Description
**This is a breaking change when compiling in Vagrant**

Allow Dobby to build in Release mode without any errors. Set the log levels appropriately.

Reduce the overall footprint of Dobby by moving libocispec to a shared library, instead of including in multiple places as a static library.

When building in the Vagrant VM, you must now use the following CMake command to ensure libocispec is built in the correct directory:
```
cmake -DCMAKE_BUILD_TYPE=Debug -DRDK_PLATFORM=DEV_VM -DCMAKE_INSTALL_PREFIX:PATH=/usr ../
```

This is handled automatically by Bitbake so only applies to the VM.

This PR also changes the default RDK platfrom from Xi6 to Generic, which is suitable for use on any general RDK platform such as Xi6 or RPi

### Test Procedure
* Attempt to build Dobby in release mode, both with and without LTO enabled on the VM and in Bitbake. No compiler errors should occur
* Build and test Dobby on various STB platforms, both in Debug and Release mode. No regressions should be visible

### Type of Change


TeknoVenus:
Splitting into smaller PRs....

PR #61, opened 2020-10-22
RDK-30110: Build in release mode without errors

TeknoVenus:
### Description
Allow Dobby to build in Release mode without any errors. Set the log levels appropriately. 

Note release builds do not include DobbyTool or DobbyBundleGenerator.

Also renames the Xi6 platform to `Generic` as Dobby Spec support is not required on Xi6 platforms so no extra settings are needed specific to Xi6. 

Legacy Components default to OFF

Adds a new CMake option for use by Bitbake to append the SRC_REV to the version number of Dobby for easy checking of build version on the STB

### Test Procedure
Build Dobby in release mode both with and without LTO enabled. Should build without errors and no regressions should be visible in Sky and generic RDK builds

### Type of Change


daemon/process/settings/dobby.json
```diff
@@ -0,0 +1,15 @@
+{
```
goruklu, line 1: @TeknoVenus  dobby.json is used only when using dobbyspec.json and not the OCI bundle, correct ?

daemon/process/CMakeLists.txt
```diff
@@ -108,8 +108,9 @@ install( # Specifies the target to install
 
 
 # Install the config file for the target platform
-if( RDK_PLATFORM STREQUAL "XI6" )
-        set( DOBBY_SETTINGS_FILE "settings/dobby.xi6.json" )
+# Generic is just a standard build with no extra settings
```
goruklu, line 111: I think we should keep the xi6 specific configuration even though it's not used in current builds.

goruklu (commented):
The changes look good to me but I'd like to keep Xi6 config in as reference in addition to the generic, Llama and Xione configurations.

daemon/process/settings/dobby.json
TeknoVenus: The `paths`, `logging` and `network` settings are used to configure the daemon, and apply for bundle and dobbyspec containers. Those settings should always be present

The remaining settings only apply to dobbyspec containers, so are not necessary on non-sky builds (hence why I removed them from the generic RDK config)

Koczek9 (dismissed):
LGTM

TeknoVenus:
@Koczek9 - Git commits went weird, but made a small change in e5b71b2 to allow enabling DobbyTool in release builds - please re-review when you get chance.

goruklu (approved):

PR #63, opened 2020-11-20
[RDK-22853] Fix mutlicastforwarder plugin

TeknoVenus:
### Description
Fixes the MulticastForwarder plugin so mutlicast traffic correctly flows between the container. This is needed for `rtRemote` to function inside a container with NAT networking.

Previously, `smcroutectl` was used to add new multicast rules. However, despite the patch to smcroute provided by @anttikyl here: https://github.com/rdkcentral/Dobby/pull/53#issuecomment-702751479, smcroute was not correctly adding the `dobby0` interface to the rules.

This fix switches to writing the rules directly to the smcroute config file then reloading the smcroute daemon, ensuring the interfaces are correctly identified and the routes are valid. This also fixes an issue where if two containers had the same multicast rule(s), then stopping one container would remove the rule and the second container would stop receiving multicast packets.

The smcroute patch is no longer needed and removed from the Vagrantfile.

### Test Procedure
`mcjoin` can be used as a mutlicast sender/receiver: https://github.com/troglobit/mcjoin to fully test. For verification, an mcjoin server was started on the dev VM and an mcjoin client run inside the container - the container correctly received the packets

* Start smcroute: `sudo /usr/local/sbin/smcrouted -n -l debug`
* Start Dobby Daemon as usual. Make sure no containers are currently running
* Create a Dobby container with the Multicast forwarder plugin enabled:

```json
"multicastForwarding": [
                    {
                        "ip": "224.10.0.12",
                        "port": 10004
                    }
                ]
```

* Check to see the wildcard route exists in smcroute:

```
ROUTE (S,G)            INBOUND        PACKETS    BYTES  OUTBOUND
(*, 224.10.0.12)       enp0s3               0        0  dobby0
```

* Start a multicast sender on the host. See that smcroute has added the actual route and packets are flowing

```
ROUTE (S,G)                 INBOUND      PACKETS      BYTES  OUTBOUND 
(10.0.2.15, 224.10.0.12)    enp0s3            46       5888  dobby0
```

* Check this route is reflected in the kernel routing table, with the correct output interface (IMPORTANT)

```
vagrant@dobby-vagrant:~$ ip mroute
(10.0.2.15, 224.10.0.12)         Iif: enp0s3     Oifs: dobby0
```

* Check the container is receiving multicast packets
* Check that the config file for smcroute is updated with the expected rules:

```
vagrant@dobby-vagrant:~$ cat /usr/local/etc/smcroute.conf 
#START:client
mroute from enp0s8 group 224.10.0.12 to dobby0
mroute from enp0s3 group 224.10.0.12 to dobby0
#END:client
```

Rules are added between two comments denoting the container ID. Starting a second container should add the rules for that container. `smcroute` will ignore duplicate entries.

```
#START:client
mroute from enp0s8 group 224.10.0.12 to dobby0
mroute from enp0s3 group 224.10.0.12 to dobby0
#END:client
#START:server
mroute from enp0s8 group 224.10.0.12 to dobby0
mroute from enp0s3 group 224.10.0.12 to dobby0
#END:server
```

* Stop the container. The rules for that container are removed from the config file

### Type of Change

The smcroute recipe needs modifying to configure smcroute to use the config file located at `/opt/smcroute.conf` instead of `/etc/smcroute.conf`, as `/etc` is readonly.


goruklu (approved):

Koczek9 (approved):
LGTM

PR #64, opened 2020-12-04
Enable multicast forwarding on localhost

TeknoVenus:
### Description
Multicast forwarder should forward packets sent on the localhost interface to `dobby0`. 

This is necessary for rtRemote which uses the localhost to send multicast packets between processes on a single host.

### Test Procedure
Same test procedure as per https://github.com/rdkcentral/Dobby/pull/63. The `lo` interface should also be added to the smcroute.conf file:

```
mroute from lo group 224.10.0.12 to dobby0
```

### Type of Change


Koczek9 (approved):
LGTM

PR #66, opened 2020-12-11
Allow enabling/disabling the start container hook

TeknoVenus:
### Description
The StartContainer hook is not used by any existing plugins and causes issues with library dependencies inside containers.

Disable the hook by default, but allow manually enabling it with a CMake flag if desired.

Whilst this does not break any existing plugin behaviour, this is a **breaking change** as the currently documented behaviour for plugin hooks is changed.

### Test Procedure
Build Dobby as normal. The `startContainer` hook should not run when starting a container and the `startContainer` section should not be populated in the container config after starting a container.

Setting `-DUSE_STARTCONTAINER_HOOK=ON` when compiling Dobby should revert this behaviour and the startContainer hook should run

### Type of Change

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,6 +185,12 @@ if( ENABLE_LTO )
     endif()
 endif()
 
+# Enable the start container hook (disabled by default)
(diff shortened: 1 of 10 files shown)


Koczek9 (approved):
LGTM

PR #67, opened 2020-12-17
Always set the container hostname to container ID

TeknoVenus:
### Description
Always set the hostname of the container to the container ID

### Test Procedure
Start a container from the same bundle multiple times, each with a different ID. The hostname of the container should always match the ID

### Type of Change

### DIFF
--- a/bundle/lib/source/DobbyConfig.cpp
+++ b/bundle/lib/source/DobbyConfig.cpp
@@ -520,6 +520,8 @@ bool DobbyConfig::setHostnameToContainerId(const ContainerId& id, std::shared_pt
     {
         return true;
     }
+
+    free(cfg->hostname);
     cfg->hostname = strdup(id.c_str());
 
     // write the new config.json to a file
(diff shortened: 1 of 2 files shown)


bundle/lib/source/DobbyConfig.cpp
```diff
@@ -505,6 +505,32 @@ void DobbyConfig::addPluginLauncherHooks(std::shared_ptr<rt_dobby_schema> cfg, c
...
+ *  @param[in]  bundlePath      path to the container bundle
+ */
+bool DobbyConfig::setHostnameToContainerId(const ContainerId& id, std::shared_ptr<rt_dobby_schema> cfg, const std::string& bundlePath)
+{
+    // change hostname to container id only if necessary
+    if (!strcmp(cfg->hostname, id.c_str()))
+    {
+        return true;
+    }
+    cfg->hostname = strdup(id.c_str());
```
Koczek9, line 523: We should use free before using strdup. Using strdup allocates new memory, meaning old pointer is lost before we used free on it. So we have memory leak here. Strange enough I have found some other uses of strdup without free in our code. We should fix those also.

Koczek9 (changes requested):
Please free allocated memory before reassigning pointer to new memory.

Koczek9 (approved):
LGTM

PR #68, opened 2021-01-07
Adjust permissions on created directories for storage plugin

TeknoVenus:
### Description
Fix issue where Storage plugin would fail to mount the img file when using a deeply nested path as the destination

### Test Procedure
Create a loopback mount with the following config:

```
"loopback": [
    {
        "destination": "/home/private/test/long/path",
        "flags": 14,
        "fstype": "ext4",
        "source": "/tmp/data/data.img"
    }
]
```

The storage plugin should run successfully. Without this PR a permission denied error will be seen.

### Type of Change


Koczek9 (approved):
LGTM

PR #69, opened 2021-01-12
Wait for container stop event before exiting DobbyTool

TeknoVenus:
### Description
Currently, when stopping a container with DobbyTool, the program will exit as soon as Dobby has sent the `crun kill` command to stop the container. However, it can take a few hundred ms for all the container processes to actually stop and for the container to exit. This can be problematic when restarting a container by doing

```
DobbyTool stop xxx && DobbyTool start
```

since the first DobbyTool command can exit before the container has completely stopped, so the subsequent start command will fail.

This PR modifies the DobbyTool stop command to wait for the `DOBBY_CTRL_EVENT_STOPPED` event before exiting. This ensures DobbyTool doesn't exit before the container has fully stopped.

### Test Procedure
Start a container with DobbyTool.

Attempt to restart the container by running:
```
DobbyTool stop xxx && DobbyTool start
```

The container should always restart successfully.

### Type of Change

### DIFF
--- a/client/tool/source/Main.cpp
+++ b/client/tool/source/Main.cpp
@@ -49,6 +49,7 @@
 #include <sys/un.h>
 #include <linux/limits.h>
 #include <dirent.h>
+#include <future>
 
 #include <list>
 #include <string>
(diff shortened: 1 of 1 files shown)


client/tool/source/Main.cpp
```diff
@@ -133,22 +157,33 @@ static void stopCommand(const std::shared_ptr<IDobbyProxy>& dobbyProxy,
         }
     }
 
+    // Register an event listener to monitor for the container stop
+    promise = std::promise<bool>();
```
Koczek9, line 161: Here there could be a problem if 2 `close` containers calls were done in very short time, as promise is global variable it could be override by second use. Furthermore `future.get` can be done only once (you would get "`std::future_error: No associated state`" if you try get more than once) - this is one more reason to use` future.wait()` for this.
We could solve this problem by creating additional mutex for promise and acquire it here (after making sure that it is valid close parameters) This way we would make sure that only one close is called at a time so global promise is safely connected to this future.

Koczek9 (changes requested):

Koczek9 (approved):
In future when you have lock for particular resource name should reflect it i.e. `gPromiseLock` but as this is only thing that I can find, and it is not harmful then no need to change it now.

LGTM

PR #70, opened 2021-01-14
RDK-30291 Fixed occasional temp mounts

Koczek9:
### Description
There was an issue where container was created and not run, it was resulting in temp mount being left because cleaning up was done in post start hook, which wasn't run as container wasn't start.
Fixed by checking if cleaning was done on post stop hook which should be run all the times.

This ticket is fix for problem found while doing RDK-30291.

### Test Procedure
Create container with storage plugin on, but without "post start" hook. It shouldn't left mount to `directory.temp`.

### Type of Change

- [x] No

### DIFF
--- a/bundle/lib/include/DobbyRootfs.h
+++ b/bundle/lib/include/DobbyRootfs.h
@@ -74,6 +74,8 @@ public:
 
     void setPersistence(bool persist);
 
+    void unmountAllAt(const std::string& pathPrefix);
+
 #if defined(LEGACY_COMPONENTS)
 private:
     void cleanUp(bool dontRemoveFiles);
@@ -90,7 +92,6 @@ private:
                               const std::string& fileContents,
                               mode_t mode = 0644) const;
 
-    void unmountAllAt(const std::string& pathPrefix);
 #endif //defined(LEGACY_COMPONENTS)
 
 private:
--- a/bundle/lib/source/DobbyRootfs.cpp
+++ b/bundle/lib/source/DobbyRootfs.cpp
@@ -145,6 +145,11 @@ DobbyRootfs::~DobbyRootfs()
 {
 #if defined(LEGACY_COMPONENTS)
     cleanUp(mPersist);
+#else
+    if (!mPath.empty())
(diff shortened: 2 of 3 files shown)


TeknoVenus:
This fixes most occurrences of this problem, but unfortunately I've found a scenario (albeit fairly niche) where this does not fix the issue.

The temp mounts are set up in the `preCreation` hook, which is run before we call `crun create`. If `crun create` fails to parse the config.json file or cannot set up the container (e.g. due to an invalid mount option), the container will not be created and no other OCI hooks will run (see https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#lifecycle).

This means the temp mounts are set up, but since no other OCI hook runs they are not cleaned up.

To reproduce:
* Build Dobby with LEGACY_COMPONENTS=OFF (important, with legacy_components=on Dobby does some cleanup in the rootfs when the container stops (in the `unmountAllAt()` function). 
* Create a container config with an invalid mount option (e.g. mount a non-existent dir into the container)
* Start container - will fail
* Check `mount` output - temp mount will still exist:
```
/tmp/data/data.img on /home/vagrant/sleepybundle/rootfs/home/private/test/123.temp type ext4 (rw,nosuid,nodev,noexec,relatime,data=ordered)
```

Possible solutions:
* Use `postHalt` hook instead to clean up the temp mount
* As mentioned earlier, when building with legacy_components, in `DobbyRootfs` there is a `unmountAllAt()` method which removes leftover mount points in the rootfs dir. Perhaps this should be enabled for all build types, which would then ensure this can never happen

Koczek9:
Remarks from longer conversations:
- We shouldn't run postHalt when container was started (we should create a ticket for this)
- Current solution of removing temp works in most cases so it should stay but we need to add additional case
- We should fix this issue by making `unmountAllAt` non-legacy component and call it in destructor of `DobbyRootfs` when `LEGACY_COMPONENTS` is not defined

TeknoVenus:
LGTM

TeknoVenus (approved):

PR #71, opened 2021-01-20
RDK-30517: Port sky dobby changes

Koczek9:
### Description
Porting changes made in sky-dobby so both repositories are in sync.

### Test Procedure
-

### Type of Change

- [x] No

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,12 +22,13 @@ cmake_minimum_required( VERSION 3.7.0 )
 include(GNUInstallDirs)
 
 # Project setup
-project( Dobby VERSION "3.0" )
+project( Dobby VERSION "3.0.0" )
 
 
 # Set the major and minor version numbers of dobby (also used by plugins)
 set( DOBBY_MAJOR_VERSION 3 )
 set( DOBBY_MINOR_VERSION 0 )
+set( DOBBY_MICRO_VERSION 0 )
 
 set(INSTALL_CMAKE_DIR lib/cmake/Dobby)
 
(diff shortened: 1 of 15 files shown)


TeknoVenus (changes requested):
The tap interface changes don't appear to work.

When testing on Xi1 with skyDobby, a `dobby0_tap` interface appears in the ifconfig output. When testing with this PR, no such interface appears in the ifconfig output on either Xi1 or Vagrant

Other than that, it looks fine.

TeknoVenus:
@Koczek9 Made a few changes to fix the tap interface issue. By default, the tap interface is destroyed when the fd is closed, which was happening when the plugin exited. Setting the tap interface to persist fixes this - it remains created until we explicitly destory it.

Will prepare Xi1 build for testing then merge.

TeknoVenus:
Testing Xi1 correctly shows the `dobby_tap0` interface:

```
# DobbyDaemon --version
Version: 3.0.0-31913a9a18817349fc5f8b032ab9fee617256da0
# ifconfig dobby_tap0
dobby_tap0 Link encap:Ethernet  HWaddr 1E:30:4C:A6:D4:32  
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```

MAC of bridge appears to be tied to tap0 correctly.

TeknoVenus (approved):

PR #76, opened 2021-02-12
RDK-31481: Destroy correct veth at container exit

TeknoVenus:
### Description
* Move `/dobbyaddress` to `/tmp/dobbyAddress_<containerid>`. This prevents it being overridden when another container is started from the same bundle
   * Add new getContainerNetworkInfo() method in utils to read/parse this file
* When launching multiple container instances from the same bundle, make sure the correct veth is destroyed on exit
* Switch to using `libocispec` to parse state from stdin instead of manually parsing JSON
    * Don't pass stdin directly to plugins - allow them to access state struct via utils clas instead

Marked as breaking due to:
* Removal of raw stdin from plugins means the plugin constructor has changed. All plugins have been updated to reflect this change. 
* /dobbyaddress file has moved, plugins depending on it need to be updated

### Test Procedure
Create a simple "sleepy" bundle with NAT networking. Start two containers from the same bundle simultaneously.

Check that the correct IP/veth is available in the `/dobbyaddress` file within the container - e.g.
```
100.64.11.2/veth0
```
```
100.64.11.3/veth1
```

When stopping one of the two containers, ensure the correct veth is destroyed and network access is not broken in the other container. When both containers are stopped, all veths should be destroyed.

### Type of Change

### DIFF
--- a/rdkPlugins/Networking/include/NetworkingHelper.h
+++ b/rdkPlugins/Networking/include/NetworkingHelper.h
@@ -23,10 +23,6 @@
 #include <netinet/in.h>
 #include <string>
 
-// TODO:: This would be better stored in the dobby workspace dir rather than /tmp,
-// but we don't programatically know the workspace dir in this code.
-#define ADDRESS_FILE_PREFIX       "/tmp/dobbyAddress_"
-
 #define BRIDGE_NAME             "dobby0"
 
 enum class NetworkType { None, Nat, Open };
--- a/rdkPlugins/Networking/include/NetworkingHelper.h
+++ b/rdkPlugins/Networking/include/NetworkingHelper.h
@@ -23,8 +23,9 @@
 #include <netinet/in.h>
 #include <string>
 
-// path to file in container rootfs where the container's ip address is stored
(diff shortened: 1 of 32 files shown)


goruklu (approved):

TeknoVenus:
@goruklu Please re-review, thanks.

rdkPlugins/Networking/include/NetworkingHelper.h
```diff
@@ -23,8 +23,9 @@
 #include <netinet/in.h>
 #include <string>
 
-// path to file in container rootfs where the container's ip address is stored
-#define ADDRESS_FILE_PATH       "/dobbyaddress"
+// TODO:: This would be better stored in the dobby workspace dir rather than /tmp,
+// but we don't programatically know the workspace dir in this code.
+#define ADDRESS_FILE_PREFIX       "/tmp/dobbyAddress_"
```
goruklu, line 28: same definition is in pluginLauncher/lib/include/DobbyRdkPluginUtils.h. Not a big deal but can we move to a common header file ?

goruklu (approved):

PR #80, opened 2021-02-19
Fix writing arrays to OCI plugin specs

kmatuszkiewicz:
### Description
When rewriting arrays from Dobby plugin specs to OCI plugin specs, source array was iterated over wrong elements, append was tried on the destination array even if it didn't exist. This resulted in invalid data being written to the OCI plugin spec.

### Test Procedure
Prepare a Dobby plugin spec which contains a non-empty array, run it through DobbyBundleGenerator and check the output config file.

### Type of Change


TeknoVenus (approved):
LGTM

PR #92, opened 2021-03-23
NGDEV-180377: Erasing loopback storage on container start

bgray-sky:
### Description
Add option to disable image management for loopback file system mounts. 
This is needed on Sky builds where the fs images are managed by appsserviced.

### Test Procedure
Build and run on a sky build, you should no longer see the following error in the log:
```
xattr missing on data file, re-generating a new one
```
And the supplied loopback image file should be used as it.

### Type of Change


TeknoVenus (approved):
LGTM!

PR #125, opened 2021-06-22
Allow to provide absolute rootfs path.

marcin-hajkowski-epam:
### Description
Add possibility to provide absolute rootfs path and thus store config.json and rootfs in different places

### Test Procedure
Provide absolute rootfs path in config.json file (https://github.com/opencontainers/runtime-spec/blob/master/config.md#root)

### Type of Change


TeknoVenus (approved):
Thanks for the contribution and sorry for the delay approving this - LGTM

PR #130, opened 2021-08-19
Add localhostmasquerade to Thunder plugin

TeknoVenus:
### Description
Add `localhostMasquerade` functionality to Thunder plugin.

When enabled, Thunder will be available at `127.0.0.1:9998` inside the container in addition to `100.64.11.1:9998`. This is useful for applications that don't correctly respect the THUNDER_ACCESS env var. 

Refactor some iptables rule generation code for re-use, as localhost masquerade feature is used across Network, AS and Thunder plugins.

### Test Procedure
When the Thunder plugin is configured with `localhostMasquerade` set:

```json
"rdkPlugins": {
        "thunder": {
            "required": true,
            "dependsOn": ["networking"],
            "data": {
                "bearerUrl": "http://localhost",
                "localhostMasquerade": true
            }
        }
    }
```
Thunder should be accessible over `127.0.0.1:9998` inside the container.

### Type of Change

### DIFF
--- a/bundle/runtime-schemas/defs-plugins.json
+++ b/bundle/runtime-schemas/defs-plugins.json
@@ -497,6 +497,9 @@
                         "bearerUrl": {
                             "type": "string"
                         },
+                        "localhostMasquerade": {
+                            "type": "boolean"
+                        },
                         "connLimit": {
                             "type": "integer"
                         }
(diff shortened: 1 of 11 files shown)


scthunderbolt:
Closed due to inactivity, please restore if still needed.

PR #141, opened 2021-09-17
Update vagrant file

TeknoVenus:
### Description
Upgrade the Vagrant VM from Ubuntu 16.04 (EOL) to Ubuntu 20.04.

Build Thunder R2 instead of Thunder master branch.

### Test Procedure
Vagrant VM should build and work correctly.

### Type of Change
- [x] Other (doesn't fit into the above categories - e.g. documentation updates)


Koczek9 (approved):
LGTM
Only remark I have is that for some reason cloning is marked with red font, and last 4 commands are clone commands. This leaves user with last 4 command which were successful but they are red, which is confusing. We could think how to make those green, or maybe write "success" as last command just so user is not confused. This doesn't need to be done now and I am fine with merging it as is if we don't have time to change it now.

TeknoVenus:
Yeah, `git clone` goes to stderr, and Vagrant makes all stderr as red.

PR #143, opened 2021-09-28
Add devicemapper plugin

TeknoVenus:
### Description
Create new dobby plugin called `DeviceMapper`

This plugin will allow Dobby to correct the major/minor IDs of specified devices in the config file, which is useful when using a pre-created bundle but some devices may dynamically change their major/minor ID each boot.

Ideally all devices should have fixed IDs, but sometimes this is not practicable.

### Test Procedure
Create a test container config which maps in a device but with an invalid major and/or minor ID. Add the devicemapper plugin to the config with the path to the device.

E.G

Container config contains the `/dev/fuse` device, with an incorrect major/minor ID
```
        "devices": [
            {
                "type": "c",
                "path": "/dev/fuse",
                "fileMode": 438,
                "major": 999,
                "minor": 999,
                "uid": 0,
                "gid": 0
            },
```

and the `deviceMapper` plugin
```
        "devicemapper": {
            "required": false,
            "data": {
                "devices": [
                    "/dev/fuse"
                ]
            }
        }
```

When starting the container, the plugin will correct the major/minor ID
```
0000017728.705885 <T-106257> NFO: < M:DobbyRdkPluginManager.cpp F:runPlugins L:756 > Running devicemapper plugin
0000017728.725337 <T-106257> NFO: < M:DeviceMapper.cpp F:preCreation L:118 > Fixing major/minor ID for dev node '/dev/fuse'
0000017728.733166 <T-106257> NFO: < M:DeviceMapper.cpp F:preCreation L:162 > Fixing major/minor ID in cgroup allow list for dev node '/dev/fuse'
0000017728.738577 <T-106257> NFO: < M:DobbyRdkPluginManager.cpp F:runPlugins L:775 > Plugin devicemapper has preCreation hook run successfully
```

Both the `devices` list and cgroup whitelist (`resouces`->`devices`) will both be corrected.

### Type of Change


rdkPlugins/DeviceMapper/README.md
```diff
@@ -0,0 +1,29 @@
+# Dobby RDK Device Mapper Plugin
+
+## Quick Start
+Add the following section to your OCI runtime configuration `config.json` file to enable logging to a file. This will capture all the stdout/err from the container (in addition to any output from `DobbyPluginLauncher`) and redirect it to the specified file.
```
Koczek9, line 4: This is copy pasted from Logging plugin. Please create proper introduction to Device Mapper plugin.

rdkPlugins/DeviceMapper/source/DeviceMapper.cpp
```diff
@@ -0,0 +1,233 @@
...
+    {
+        if (getDevNodeFromPath(mContainerConfig->rdk_plugins->devicemapper->data->devices[i], node))
+        {
+            deviceNodes.emplace_back(node);
+        }
+    }
+
+    // We know we have some devices we're interested in, but we didn't find any
+    // of them, so something's gone wrong
+    if (deviceNodes.size() == 0)
```
Koczek9, line 94: So why aren't we also comparing 
`deviceNodes.size() != mContainerConfig->rdk_plugins->devicemapper->data->devices_len`?
Shouldn't them being not equal means that something went wrong as well? I know it gets logged as warning when some device was inaccessible which is good, but if it is fine when at least one device was mapped we should probably write some comment in README that this is the case. Or maybe we could add additional flag like "mandatory" which if true would mean that all devices must be mapped, and set it default false.

rdkPlugins/DeviceMapper/source/DeviceMapper.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+        if (it == incorrectDevNodes.end())
+        {
+            continue;
+        }
+
+        AI_LOG_INFO("Fixing major/minor ID in cgroup allow list for dev node '%s'", it->path.c_str());
+        configDev->major = it->major;
+        configDev->minor = it->minor;
+    }
```
Koczek9, line 174: We are not checking if we handled every `incorrectDevNodes` entry. Maybe we should delete entry after fixing it and, after this whole loop, check for `incorrectDevNodes.size() != 0` again. It would mean someone had some issue with misspelling either device path or id, but it would have also the benefit of speeding computation in case of long lists of device to be mapped (which we both know no one will ever do *wink-wink*).

Koczek9 (approved):
LGTM

PR #147, opened 2021-10-29
XIONE-6236 Don't wait for failed forks

Koczek9:
### Description
We shouldn't do waitpid if fork failed, as if there was another child it will result in infinite wait. Also why do we even use waitpid at all? In forkExecRunC there is implementation based on vfork which is blocking by definition. Which should mean that if we got to this line we already finished the child process. I guess it wouldn't harm (at least as long as we check if it succeed to fork on the first place).

### Test Procedure
We would somehow need to fail the fork, while having other child process. Only disabling fork at all would get "10 - No child processes" result.

### Type of Change

- [x] No


TeknoVenus:
As discussed, this change is safe to merge - further enhancements in this area are a WIP

TeknoVenus (approved):

PR #154, opened 2021-11-17
Don't call ip6tables if ipv6 disabled in config

TeknoVenus:
### Description
If IPv6 is disabled in the container config, Dobby should never attempt to call `ip6tables`.

### Test Procedure
Create container config with ipv6 disabled. Launch container. Dobby should never attempt to run `ip6tables-restore`. Dobby should only run `ip6tables-restore` when `ipv6: true` is set in the networking config.

Needed on VAs where ip6tables-restore isn't working (ipv6 support not needed)

### Type of Change

### DIFF
--- a/rdkPlugins/Networking/source/Netfilter.cpp
+++ b/rdkPlugins/Networking/source/Netfilter.cpp
@@ -484,6 +484,8 @@ bool Netfilter::applyRules(const int ipVersion)
 {
     AI_LOG_FN_ENTRY();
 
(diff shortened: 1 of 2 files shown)


Koczek9 (approved):
LGTM

TeknoVenus:
Moved back to draft for the time being - requires further testing.

There are some other areas where ipv6 is configured even if not set in plugin config, need to review code further

scthunderbolt:
Closed due to inactivity, please restore if still needed.

PR #156, opened 2021-11-23
RDKTV-12390: Handle long container log lines better with  journald

TeknoVenus:
### Description
Prevent Dobby from locking up and using excessive CPU when printing very long log messages to journald.

This only impacts the journald log sink, logging to files was not affected by this issue

### Test Procedure
Create test container that prints a very long single line (>8000 chars) and configure logging plugin to use journald.

Log message should be printed to journald without error. Note very, very long messages max be broken up into multiple lines based on the journal's LineMax setting

### Type of Change

### DIFF
--- a/rdkPlugins/Logging/source/LoggingPlugin.cpp
+++ b/rdkPlugins/Logging/source/LoggingPlugin.cpp
@@ -272,12 +272,18 @@ void LoggingPlugin::JournaldSink(const ContainerInfo &containerInfo,
     memset(buf, 0, sizeof(buf));
 
     ssize_t ret;
-    ssize_t offset = 0;
 
     // Create a file descriptor we can write to
-    // journald will handle line breaks etc automatically so we don't need
-    // to worry about them
+    // journald will handle line breaks etc automatically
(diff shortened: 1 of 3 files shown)


rdkPlugins/Logging/source/LoggingPlugin.cpp
```diff
@@ -274,70 +268,45 @@ void LoggingPlugin::JournaldSink(const ContainerInfo &containerInfo, bool exitEo
...
-                                "PRIORITY=%i", logPriority,
-                                "SYSLOG_IDENTIFIER=%s", mUtils->getContainerId().c_str(),
-                                "OBJECT_PID=%ld", containerInfo.containerPid,
-                                "SYSLOG_PID=%ld", containerInfo.containerPid,
-                                NULL);
-            }
-
-            lineStart = lineEnd + 1;
-        }
+        offset += ret;
```
Koczek9, line 295: This offset variable seams unused. What should it do?

TeknoVenus: not needed anymore, leftover from old implementation. Removed

Koczek9 (approved):
LGTM

PR #159, opened 2021-12-08
XIONE-7154 Plugin launcher hook execution timeout

Koczek9:
### Description
Added timeout for hook execution in plugin launcher because it looks like some plugins freezes. With additional log provided by this commit we should be able to pinpoint the offender.

### Test Procedure
Modify any plugin to have some long (more than 4 seconds) wait in it. This should make timeout happen and if plugin is required then container will fail to start.

### Type of Change

### DIFF
--- a/pluginLauncher/lib/source/DobbyRdkPluginManager.cpp
+++ b/pluginLauncher/lib/source/DobbyRdkPluginManager.cpp
@@ -34,6 +34,9 @@
 #include <dlfcn.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
+#include <sys/mman.h>
+#include <signal.h>
 
 #include <functional>
 #include <list>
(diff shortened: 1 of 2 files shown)


pluginLauncher/lib/source/DobbyRdkPluginManager.cpp
```diff
@@ -659,25 +662,131 @@ bool DobbyRdkPluginManager::executeHook(const std::string &pluginName,
 
 // -----------------------------------------------------------------------------
 /**
- * @brief Run the plugins specified in the container config at the given hook point.
- * Returns true if all required plugins execute successfully. If non-required plugins
- * fail or are not loaded, then it logs an error but continues running other plugins
- *
- * @param[in]   hookPoint   Which hook point to execute
+ * Runs the specified hook for a given plugin, controls if execution takes
```
TeknoVenus, line 665: Not quite sure what you mean by `controls` here?

TeknoVenus (changes requested):

TeknoVenus (approved):
LGTM - in testing adding sleep in plugins correctly triggers a timeout :+1:

PR #164, opened 2022-01-11
RDKTV-13083 Get backup config if current one is corrupted

Koczek9:
### Description
In case problems with parsing config try to get backup and run container from it.

### Test Procedure
Run any container once (to make sure backup is created). Stop container and then remove some content from config.json so it is no longer valid (i.e. any brace symbol). Try to run container again, with this fix everything should work fine

### Type of Change


TeknoVenus (changes requested):
* You've only done this change for DobbySpecConfig - not DobbyBundleConfig so this won't apply when loading from OCI bundles (main use case here)

* If we restore from backup, we will need to run the postInstallation hooks again. Remove the `postinstallsuccess` flag if it exists to ensure this occurs

TeknoVenus (approved):

PR #165, opened 2022-01-11
RDKTV-13083 Cleanup config is container failed to start

Koczek9:
### Description
Cleaning up temp config in case container failed to start.

### Test Procedure
 Start container with custom argument (e.g. DobbyTool start sleep echo "hello"), Dobby will create a custom config.json file called config-XXX.json, where XXX is the container descriptor (see https://github.com/rdkcentral/Dobby/blob/master/daemon/lib/source/DobbyManager.cpp#L452). With this fix if container fails to start then custom config should be deleted. Without fix if container fails to start custom config will stay forever.

### Type of Change


goruklu (approved):

TeknoVenus (approved):
LGTM

PR #169, opened 2022-01-25
XIONE-7601: Remove IPCService from Networking Plugin

TeknoVenus:
### Description
In some scenarios, we've seen IPCService hang when running the Networking plugin createRuntime/postHalt hooks.

To completely resolve this issue, this PR removes all IPCService usage from the Networking plugin. Previously the daemon maintained a pool of IP addresses to allocate to containers, which was the primary need for daemon <-> plugin IPC. 

This PR creates a new disk-backed IPAllocator which does not require communication with the main daemon.

### Test Procedure
Requires extensive testing of the following features:

* Ensure containers with NAT networking start without errors
* Ensure veth devices are created/destroyed correctly
* Ensure the dobby0 bridge is created when the first container starts, and removed when the final container stops
* Ensure each container has a unique IP address and the iptables rules shown in `iptables-save` are accurate
* Ensure port forwarding works
* Ensure plugins which depend on Networking (e.g. Thunder and AS) continue to work correctly

### Type of Change


pluginLauncher/lib/include/DobbyRdkPluginUtils.h
```diff
@@ -39,15 +39,28 @@
...
 typedef struct ContainerNetworkInfo
 {
     std::string vethName;
     std::string ipAddress;
+    in_addr_t ipAddressRaw;
+    std::string containerId;
+
+    bool operator==(const ContainerNetworkInfo &rhs) const
+    {
+        if (containerId.empty())
```
Koczek9, line 58: The condition should probably be:
if (containerId.empty() || rhs.containerId.empty())
because as for now there are situations where A==B and B==A could have different results.

pluginLauncher/lib/source/DobbyRdkPluginUtils.cpp
```diff
@@ -146,44 +153,31 @@ std::string DobbyRdkPluginUtils::getContainerId() const
...
+    if (addressFileStr.length() <= ipStr.length() + 1)
     {
-        AI_LOG_ERROR("failed to get veth name from %s", fileName.c_str());
+        AI_LOG_ERROR("failed to get veth name from %s", filePath.c_str());
         return false;
     }
 
-    networkInfo.ipAddress = ip;
-    networkInfo.vethName = addressFileStr.substr(ip.length() + 1, addressFileStr.length());
+    networkInfo.containerId = basename(filePath.c_str());
```
Koczek9, line 176: If you would move this line to line 157, then you could use:
networkInfo.containerId = getContainerId();
std::string filePath = ADDRESS_FILE_DIR + networkInfo.containerId;
This would mean one less basename function.

pluginLauncher/lib/source/DobbyRdkPluginUtils.cpp
```diff
@@ -146,44 +153,31 @@ std::string DobbyRdkPluginUtils::getContainerId() const
...
-                          "container from %s",
-                          fileName.c_str());
+        AI_LOG_ERROR_EXIT("failed to get IP address and veth name assigned to container from %s",
+                          filePath.c_str());
         return false;
     }
 
-    const std::string ip = addressFileStr.substr(0, addressFileStr.find("/"));
+    const std::string ipStr = addressFileStr.substr(0, addressFileStr.find("/"));
+    const in_addr_t ip = std::stoi(ipStr);
```
Koczek9, line 167: If you move this after condition then we don't need to call stoi it for invalid veth.

pluginLauncher/lib/source/DobbyRdkPluginUtils.cpp
```diff
@@ -146,44 +153,31 @@ std::string DobbyRdkPluginUtils::getContainerId() const
...
-        AI_LOG_ERROR("failed to get veth name from %s", fileName.c_str());
+        AI_LOG_ERROR("failed to get veth name from %s", filePath.c_str());
         return false;
     }
 
-    networkInfo.ipAddress = ip;
-    networkInfo.vethName = addressFileStr.substr(ip.length() + 1, addressFileStr.length());
+    networkInfo.containerId = basename(filePath.c_str());
+    networkInfo.ipAddressRaw = ip;
+    networkInfo.ipAddress = ipAddressToString(htonl(ip));
```
Koczek9, line 178: I thought that you just had ipStr with sting version IP address, if this is because you need special formatting, then fine if not then just use ipStr here. Same for getNetworkInfo.

TeknoVenus:
PR re-opened here: https://github.com/rdkcentral/Dobby/pull/170

rdkPlugins/Networking/source/IPAllocator.cpp
```diff
@@ -0,0 +1,315 @@
...
+    if (!updateFromStore())
+    {
+        AI_LOG_ERROR("Failed to initialise IP backing store");
+    }
+
+    // Start from xxx.xxx.xxx.2 to leave xxx.xxx.xxx.1 open for bridge device
+    in_addr_t addrBegin = INADDR_BRIDGE + 1;
+    in_addr_t addrEnd = addrBegin + TOTAL_ADDRESS_POOL_SIZE;
+
+    // Populate the pool of available addresses
```
Koczek9, line 45: Should something be here? This newline suggest to me that you wanted to add some piece of code here.

rdkPlugins/Networking/source/IPAllocator.cpp
```diff
@@ -0,0 +1,315 @@
...
+    AI_LOG_FN_EXIT();
+    return true;
+}
+
+/**
+ * @brief Synchronise the in-memory pool of allocated IPs with the disk store
+ *
+ * @return True on success
+ */
+bool IPAllocator::updateFromStore()
```
Koczek9, line 223: I don't like how this method looks like. You are not clearing mAllocatedIps on beginning and using emplace_back which suggest that you expect that someone can run "updateFromStrore" while Allocator is already running, but at the same time you are not updating mUnallocatedIps. So I am not sure what is the desired efect here. If you would like to have updates while running please update also the mUnallocatedIps (but this would make this so much like the constructor itself). If this is helper function for the constructor and shouldn't be running after Allocator was constructed then clean mAllocatedIps at the beggining (in case someone us it like that) and maybe change the name of this method.

rdkPlugins/Networking/source/NetworkSetup.cpp
```diff
@@ -1030,6 +1014,8 @@ void NetworkSetup::addResolvMount(const std::shared_ptr<DobbyRdkPluginUtils> &ut
     const std::string source = "/etc/resolv.conf";
     const std::string destination = "/etc/resolv.conf";
 
+    AI_LOG_INFO("HERE3");
```
Koczek9, line 1017: Remove debug prints, or change to some valid message with debug priority.

rdkPlugins/Networking/source/NetworkSetup.cpp
```diff
@@ -1041,6 +1027,8 @@ void NetworkSetup::addResolvMount(const std::shared_ptr<DobbyRdkPluginUtils> &ut
         }
     }
 
+    AI_LOG_INFO("HERE4");
```
Koczek9, line 1030: Remove debug prints, or change to some valid message with debug priority.

rdkPlugins/Networking/source/NetworkingPlugin.cpp
```diff
@@ -127,6 +124,7 @@ bool NetworkingPlugin::postInstallation()
         // a new /etc/resolv.conf is created rather than mounting the host's
         if (!mPluginData->dnsmasq)
         {
+            AI_LOG_INFO("Adding resolv.conf mount");
```
Koczek9, line 127: Maybe change to debug priority.

Koczek9 (changes requested):

PR #170, opened 2022-01-25
XIONE-7601: Remove IPCService from Networking Plugin

TeknoVenus:
### Description
In some scenarios, we've seen IPCService hang when running the Networking plugin createRuntime/postHalt hooks.

To completely resolve this issue, this PR removes all IPCService usage from the Networking plugin. Previously the daemon maintained a pool of IP addresses to allocate to containers, which was the primary need for daemon <-> plugin IPC. 

This PR creates a new disk-backed IPAllocator which does not require communication with the main daemon.

### Test Procedure
Requires extensive testing of the following features:

* Ensure containers with NAT networking start without errors
* Ensure veth devices are created/destroyed correctly
* Ensure the dobby0 bridge is created when the first container starts, and removed when the final container stops
* Ensure each container has a unique IP address and the iptables rules shown in `iptables-save` are accurate
* Ensure port forwarding works
* Ensure plugins which depend on Networking (e.g. Thunder and AS) continue to work correctly

### Type of Change


Koczek9 (approved):
LGTM, the only thing is ipAddressToString is defined in two different places with same body, but we can live with it.

PR #171, opened 2022-01-31
RDKTV-12717/RDKTV-13079: Support relaying syslog/journald messages

TeknoVenus:
### Description
When containers attempt to log directly to journald (e.g. `sd_journal_print()`) or syslog, the log messages may not correctly appear in journald.

After adding `/run/systemd/journal/socket` into a container that directly calls `sd_journal_print`, the log messages are seen in `journalctl -f`. However, these messages are not tagged with a `_SYSTEMD_UNIT` tag. This is because the message unit is determined using the PID of the process, which does not match with a known unit PID when the process runs in a container.

As a result, when filtering the logs by unit (`journalctl -fu dobby`), the messages do not appear. This is an issue with RDK log collection scripts.

This PR makes the following changes:
* Based on the relevant setting in `/etc/dobby.json`, create dobby syslog and journald sockets that can be mounted into a container instead. When messages are received on these sockets, Dobby will forward the messages to the "real" socket on the host
    * Since Dobby sends the message to journald, the PID resolves to DobbyDaemon and the messages can be seen when doing `journalctl -fu dobby`
* Re-write much of Dobby logging code to use an epoll based implementation instead of spinning up a thread per-container
    * Since log relays are using epoll, it made sense to standardise all the logging methods to epoll at the same time
    * Improves performance and helps prevent lock-up during shutdown whilst we wait for loggers to finish
    * Observed significantly reduced CPU usage from DobbyDaemon during periods of high log output

### Test Procedure
**This is a high-risk change due to amount of code changed - please test thoroughly**

Create a simple C program with the following contents

```c
#include <systemd/sd-journal.h>
#include <syslog.h>

#include <stdio.h>
#include <unistd.h>

int main(int argc, char const *argv[])
{
    for (;;)
    {
        fprintf(stdout, "Hello world - stdout test!\n");
        fprintf(stderr, "Hello world - stderr test\n");
        syslog(LOG_INFO, "Hello world, this is a syslog test");
        sd_journal_print(LOG_INFO, "Hello world, this is a journald test");

        sleep(1);
    }

    return 0;
}
```

Create a container config to run this program.

Do the following (in Vagrant):
* Start Dobby as systemd service
* Add `/dev/log` and `/run/systemd/journal/socket` mounts to the container. 
    * Run `journalctl -f`. Observe the syslog and journald messages
    * Run `journalctl -f -u dobby`. Observe the syslog and journald messages **do not appear**

* Remove previous bind-mounts. Add `/tmp/dobby-syslog` and `/tmp/dobby-journald` mounts to the container. 
    * Run `journalctl -f`. Observe the syslog and journald messages
    * Run `journalctl -f -u dobby`. Observe the syslog and journald messages **do appear**

Due to the changes elsewhere to logging code, also test the following features:
* Logging to file works correctly
    * Ensure logs are correctly written even if container fails to start
    * Ensure log files are correctly re-written each time container starts (instead of appending)
* Logging to journald correctly
    * When journald logging enabled with container, with the test code provided earlier, observe all 4 messages appearing in journald
* Logging to devnull sink works correctly (e.g. no logs printed)
* Ensure Dobby shuts down without error

### Type of Change

### DIFF
--- a/daemon/lib/source/DobbyLogRelay.cpp
+++ b/daemon/lib/source/DobbyLogRelay.cpp
@@ -1,3 +1,22 @@
+/*
+* If not stated otherwise in this file or this component's LICENSE file the
+* following copyright and licenses apply:
+*
+* Copyright 2022 Sky UK
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
 #include "DobbyLogRelay.h"
 
 #include <unistd.h>
@@ -12,7 +31,8 @@
 DobbyLogRelay::DobbyLogRelay(const std::string &sourceSocketPath,
                              const std::string &destinationSocketPath)
     : mSourceSocketPath(sourceSocketPath),
-      mDestinationSocketPath(destinationSocketPath)
+      mDestinationSocketPath(destinationSocketPath),
+      mBuf{}
 {
     AI_LOG_FN_ENTRY();
 
@@ -74,17 +94,16 @@ void DobbyLogRelay::process(const std::shared_ptr<AICommon::IPollLoop> &pollLoop
 {
     if (events & EPOLLIN)
     {
-        char buf[BUFFER_SIZE] = {};
         ssize_t ret;
+        memset(mBuf, 0, sizeof(mBuf));
 
-        ret = TEMP_FAILURE_RETRY(read(mSourceSocketFd, buf, sizeof(buf)));
-
+        ret = TEMP_FAILURE_RETRY(read(mSourceSocketFd, mBuf, sizeof(mBuf)));
         if (ret < 0)
         {
             AI_LOG_SYS_ERROR(errno, "Errror during read");
         }
 
-        if (sendto(mDestinationSocketFd, buf, sizeof(buf), 0, (struct sockaddr *)&mDestinationSocketAddress, sizeof(mDestinationSocketAddress)) < 0)
+        if (sendto(mDestinationSocketFd, mBuf, sizeof(mBuf), 0, (struct sockaddr *)&mDestinationSocketAddress, sizeof(mDestinationSocketAddress)) < 0)
         {
             AI_LOG_SYS_ERROR(errno, "sento failed");
         }
--- a/daemon/lib/source/DobbyLogRelay.cpp
+++ b/daemon/lib/source/DobbyLogRelay.cpp
@@ -9,21 +9,65 @@
 
 #include <Logging.h>
 
-DobbyLogRelay::DobbyLogRelay(const int sourceSocketFd, const std::string &destinationSocketPath)
-      : mSourceSocketFd(sourceSocketFd),
+DobbyLogRelay::DobbyLogRelay(const std::string &sourceSocketPath,
+                             const std::string &destinationSocketPath)
+    : mSourceSocketPath(sourceSocketPath),
       mDestinationSocketPath(destinationSocketPath)
 {
     AI_LOG_FN_ENTRY();
 
-    // Set up the socket we will relay messages to
-    mDestinationSocketAddress = {};
-    mDestinationSocketFd = socket(AF_UNIX, SOCK_DGRAM, 0);
-    mDestinationSocketAddress.sun_family = AF_UNIX;
-    strcpy(mDestinationSocketAddress.sun_path, mDestinationSocketPath.c_str());
+    mSourceSocketFd = createDgramSocket(mSourceSocketPath);
+    if (mSourceSocketFd < 0)
+    {
+        AI_LOG_ERROR("Failed to create socket at %s", sourceSocketPath.c_str());
+    }
 
-    AI_LOG_INFO("Created log relay from fd %d to %d", sourceSocketFd, mDestinationSocketFd);
+    if (access(mDestinationSocketPath.c_str(), F_OK) < 0)
+    {
+        AI_LOG_ERROR("Socket %s does not exist, cannot create relay", mDestinationSocketPath.c_str());
+    }
+    else
+    {
+        // Connect to the socket we will relay messages to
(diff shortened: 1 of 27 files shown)


rdkPlugins/Logging/source/FileSink.cpp
```diff
@@ -0,0 +1,282 @@
...
+            // Write to the output file
+            if (write(mOutputFileFd, mBuf, ret) < 0)
+            {
+                AI_LOG_SYS_ERROR(errno, "Write failed");
+            }
+        }
+        else
+        {
+            // Hit the limit, send the data into the void
+            if (!mLimitHit)
```
Koczek9, line 155: You hit the limit and wrote to the dev/null one run before. I would get rid of this condition and move this message below to line 161. With current implementation user can lost exactly one last packet of logs and not get notified which can cause confusion.

rdkPlugins/Logging/source/FileSink.cpp
```diff
@@ -0,0 +1,282 @@
...
+            }
+            mLimitHit = true;
+            write(mDevNullFd, mBuf, ret);
+        }
+    }
+
+#if (AI_BUILD_TYPE == AI_DEBUG)
+    // Separate sections of log file for reabability
+    // (useful if we're writing lots of buffer dumps)
+    std::string marker = "---------------------------------------------\n";
```
Koczek9, line 168: Here checking for mLimitHit would make sense, as if we already exceeded limit then probably we don't wont to write anything more.

rdkPlugins/Logging/source/FileSink.cpp
```diff
@@ -0,0 +1,282 @@
...
+                // Write to the output file
+                if (write(mOutputFileFd, mBuf, ret) < 0)
+                {
+                    AI_LOG_SYS_ERROR(errno, "Write to %s failed", mOutputFilePath.c_str());
+                }
+            }
+            else
+            {
+                // Hit the limit, send the data into the void
+                if (!mLimitHit)
```
Koczek9, line 217: You hit the limit and wrote to the dev/null one run before. I would get rid of this condition and move this message below to line 223. With current implementation user can lost exactly one last packet of logs and not get notified which can cause confusion.

rdkPlugins/Logging/source/JournaldSink.cpp
```diff
@@ -0,0 +1,179 @@
...
+    std::lock_guard<std::mutex> locker(mLock);
+    mLoggingOptions = options;
+}
+
+void JournaldSink::DumpLog(const int bufferFd)
+{
+    std::lock_guard<std::mutex> locker(mLock);
+
+    ssize_t ret;
+    memset(mBuf, 0, sizeof(mBuf));
```
Koczek9, line 106: Small optimization - memset doesn't need to be part of critical section. You can move mutex below this line, so buffer can be acquired even while other thread is holding the mutex.

rdkPlugins/Logging/source/JournaldSink.cpp
```diff
@@ -0,0 +1,179 @@
...
+            {
+                AI_LOG_SYS_ERROR(errno, "Write to journald stream failed");
+            }
+        }
+
+        return;
+    }
+
+    // Container shutdown
+    if (events & EPOLLHUP)
```
Koczek9, line 156: There is no way to get to this point if previous is true, I would add use "else if" for clarity.

rdkPlugins/Logging/source/NullSink.cpp
```diff
@@ -0,0 +1,144 @@
...
+    AI_LOG_FN_EXIT();
+}
+
+NullSink::~NullSink()
+{
+    if (mDevNullFd > 0)
+    {
+        if (close(mDevNullFd) < 0)
+        {
+            AI_LOG_SYS_ERROR(errno, "Failed to close journald stream");
```
Koczek9, line 54: Message should be more like "Failed to close /dev/null stream"

rdkPlugins/Logging/source/NullSink.cpp
```diff
@@ -0,0 +1,144 @@
...
+    std::lock_guard<std::mutex> locker(mLock);
+    mLoggingOptions = options;
+}
+
+void NullSink::DumpLog(const int bufferFd)
+{
+    std::lock_guard<std::mutex> locker(mLock);
+
+    ssize_t ret;
+    memset(mBuf, 0, sizeof(mBuf));
```
Koczek9, line 70: Small optimization - memset doesn't need to be part of critical section. You can move mutex below this line, so buffer can be acquired even while other thread is holding the mutex.

rdkPlugins/Logging/source/NullSink.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            AI_LOG_SYS_ERROR(errno, "Failed to close container connection");
+        }
+
+        return;
+    }
+
+    // Don't handle any other events
+    return;
+}
```
Koczek9, line 144: I feel like we could make some one common interface for NullSink.cpp and JournaldSink.cpp to avoid copy-pasting same code. But as there are only 2 of them I won't be insisting.

Koczek9 (changes requested):

daemon/lib/source/DobbyLogRelay.cpp
```diff
@@ -0,0 +1,200 @@
...
+ * at the given path if it exists.
+ *
+ * @param[in]   path    The path the socket should be created at
+ */
+int DobbyLogRelay::createDgramSocket(const std::string &path)
+{
+    AI_LOG_FN_ENTRY();
+
+    // Remove the socket if it exists already...
+    unlink(path.c_str());
```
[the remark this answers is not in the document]
TeknoVenus: Added log

rdkPlugins/Logging/LogRelay.md
```diff
@@ -0,0 +1,53 @@
...
+        "socketPath": "/tmp/dobby-syslog"
+      },
+      "journald": {
+        "enable": true,
+        "socketPath": "/tmp/dobby-journald"
+      }
+    }
+```
+
+Then add bind mounts for the dobby sockets into the container. E.G
```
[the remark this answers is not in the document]
TeknoVenus: As discussed, this can be addressed in a future PR to avoid making this PR even more complicated

rdkPlugins/Logging/source/FileSink.cpp
TeknoVenus: Done

rdkPlugins/Logging/source/JournaldSink.cpp
TeknoVenus: Done

Koczek9 (approved):
LGTM

PR #173, opened 2022-02-09
XIONE-7775: Handle containers that don't respond to SIGKILL during cleanup

TeknoVenus:
### Description
In some scenarios on Xi1, processes can end up in a state where they don't respond to SIGKILL (assumed uninterruptible sleep).

In this case, Dobby may get stuck in a crash loop when attempting to cleanup the container. 

The following logic is introduced:
* At Dobby start, check for leftover containers
    * If container is in Running state, attempt to kill with SIGKILL
        * If SIGKILL success - delete container
        * If SIGKILL fail - mark container as "stuck" and do not attempt to delete
           * Add the container to our internal list of running containers in the Unknown state to prevent the container from being restarted
     * If container in stopped state, attempt to delete
         * If delete fails, then mark container as "stuck"
* If we have any containers marked as "stuck", regularly check if they are still stuck. Kill/delete and remove from our list if/when they eventually die.

### Test Procedure
Simulate container in uninterruptible sleep as per reproduction steps here: https://github.com/containers/crun/issues/868

Stop and restart dobby daemon. Dobby should handle the issue as per logic outlined above.

Test on STB and ensure no regressions

### Type of Change

### DIFF
--- a/daemon/lib/source/DobbyContainer.h
+++ b/daemon/lib/source/DobbyContainer.h
@@ -86,7 +86,7 @@ public:
 public:
     pid_t containerPid;
     bool hasCurseOfDeath;
-    enum class State { Starting, Running, Stopping, Paused } state;
+    enum class State { Starting, Running, Stopping, Paused, Unknown } state;
     std::string customConfigFilePath;
 
 public:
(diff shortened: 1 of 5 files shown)


Koczek9 (dismissed):
LGTM, just remember to merge it together with crun fix :). Tested on Vagrant.

Koczek9 (approved):

TeknoVenus:
Holding off merge until https://github.com/containers/crun/pull/869 is merged

With crun PR applied, Dobby will not hang during container destroy

PR #174, opened 2022-02-15
XIONE-7775: Fix cleanup loop with multiple containers

TeknoVenus:
### Description
Ensure we don't break out of the cleanupContainers loop too early.

Fix regression introduced by https://github.com/rdkcentral/Dobby/pull/173

### Test Procedure
Start multiple containers. Stop DobbyDaemon. Restart DobbyDaemon. All containers should be cleaned up

### Type of Change


Koczek9 (approved):
LGTM

PR #175, opened 2022-02-15
Fix spurious error messages

TeknoVenus:
### Description
Fix some spurious error messages seen on STB due to uninitialised variables.

Ensure variables have correct default values instead of whatever randomly gets assigned to them.

### Test Procedure
Check the following spurious errors aren't present during startup or shutdown
```
failed to find timer with id 1952541797 to remove
```
```
Syslog relay enabled but no socket path set in settings
Journald relay enabled but no socket path set in settings
```

### Type of Change


Koczek9 (approved):
LGTM
Although you could also set syslogSocketPath to empty string.

PR #179, opened 2022-02-25
LLAMA-5460: Ensure iptables rules are cleaned up after Dobby restart

TeknoVenus:
### Description
Currently when restarting Dobby, containers are cleaned up but the postHalt plugin is not run. This can result in leftover iptables rules which can cause conflicts when starting new containers.

When Dobby is shut down gracefully, call StopContainer to correctly stop containers.

If containers are detected after a dobby restart (e.g. if daemon did not gracefully shut down, perhaps after a crash), then make a best-effort attempt to run the postHalt hook.

This will impact shutdown/startup performance by 1-2 seconds if many containers were running, but ensures that the device is left in a good state so is worth the impact.

### Test Procedure
Test both graceful (SIGTERM) and force (SIGKILL) shutdown of Dobby daemon.

* Networking should work correctly when launching the same container after daemon restart
* Dobby iptables rules displayed in `iptables-save` should be removed when cleaning up containers

### Type of Change

### DIFF
--- a/daemon/lib/source/DobbyManager.cpp
+++ b/daemon/lib/source/DobbyManager.cpp
@@ -261,8 +261,9 @@ void DobbyManager::setupWorkspace(const std::shared_ptr<IDobbyEnv> &env)
 /**
  *  @brief Gets a list of running containers and tries to kill and delete them.
  *
- *  This is used at start-up so we know we have no containers left over from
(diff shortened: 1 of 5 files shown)


Koczek9 (approved):
LGTM

PR #181, opened 2022-03-14
LLAMA-5470: PollLoop sources should unregister themselves

TeknoVenus:
### Description
Ensure that all pollLoop sources unregister themselves cleanly instead of just attempting to unregister all sources after the fact.

### Test Procedure
Ensure no segfault seen during DobbyDaemon shutdown

### Type of Change


Koczek9 (commented):
You can get rid of `delAllSources` now as it is unused.

Koczek9 (approved):
LGTM

PR #183, opened 2022-03-16
Switch to Type=notify systemd service instead of Type=dbus

TeknoVenus:
### Description
Change Dobby service from `Type=dbus` to `Type=notify` to ensure that systemd does not mark Dobby service as ready too early.

Update service file with following changes:
* Switch to notify service type
* Increase watchdog timeout from 10 seconds -> 30 seconds to help prevent DobbyDaemon being killed during bootup on some platforms
* Add `Before=wpeframework` condition to ensure Dobby starts before Thunder and shuts down afterwards
   * Note on platforms without Thunder this will not cause an issue as systemd will just ignore the condition

### Test Procedure
Ensure systemd reports `Started RDK Dobby (Container) daemon.` only when Dobby is completely ready

### Type of Change

### DIFF
--- a/daemon/lib/source/Dobby.cpp
+++ b/daemon/lib/source/Dobby.cpp
@@ -552,7 +552,7 @@ void Dobby::run() const
     ret = sd_notify(0, "STOPPING=1");
     if (ret < 0)
     {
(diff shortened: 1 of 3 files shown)


Koczek9 (approved):
LGTM

PR #184, opened 2022-03-22
LLAMA-5905: Implemented a new 'bind-loop' mount type for shared private storage b…

jignatius:
…etween parent and child apps

### Description
This feature allows shared private storage for apps. This involves creating one loop device for a data.img file and bind mounting that into the containers of the different apps.

### Test Procedure
- Make the Dobby daemon and install it on the vagrant development VM:
  $ `cmake -DCMAKE_BUILD_TYPE=Debug -DRDK_PLATFORM=DEV_VM -DLEGACY_COMPONENTS=ON -DCMAKE_INSTALL_PREFIX:PATH=/usr ../`
  $ `sudo make install`
- Run the Dobby daemon in one terminal:
  $ `sudo /usr/sbin/DobbyDaemon --nofork`
- Create a test config.json. Example:
  `{
  "version": "1.1",
  "args": [
    "/bin/sleep", "60"
  ],
  "cwd": "/home",
  "console": {
    "limit": 1048576,
    "path": "/tmp/console.log"
  },
  "env": [
  ],
  "etc": {
    "hosts": [
      "127.0.0.1\tlocalhost"
    ]
  },
  "memLimit": 498073600,
  "network": "open",
  "rdkPlugins": {
  },
  "restartOnCrash": false,
  "user": {
    "gid": 30000,
    "uid": 30015
  },
  "mounts": [
          {
             "destination": "/home/private",
             "fstype": "ext4",
             "options": ["nosuid", "nodev", "noexec" ],
             "source": "/home/vagrant/sleepy_test/data.img",
             "type": "bind-loop"
        }
   ]
}`
- Launch an app using this config.json:
  $ `/usr/bin/DobbyTool start sleepy1 config.json`
- Wait 20-30s and launch another app:
  $ `/usr/bin/DobbyTool start sleepy2 config.json`
- Check only one loop device is created for both apps:
  $ `losetup -a`
- Repeat the last step to check the device loop is released only after the second app has finished running.

### Type of Change


rdkPlugins/Storage/source/StorageHelper.cpp
```diff
@@ -407,6 +407,33 @@ void StorageHelper::cleanMountLostAndFound(const std::string& mountPoint,
...
+ *  @param[in] backingFile   Path of backing file.
+ *
+ *  @return on success a positive file desccriptor corresponding to a free
+ *  loop device, -1 on error.
+ */
+std::string StorageHelper::getLoopDevice(const std::string &backingFile)
+{
+    AI_LOG_FN_ENTRY();
+    
+    const std::string command = "losetup --associated " + backingFile;
```
bgray-sky, line 423: I personally don't like calling out to userspace tools for this, it's prone too breakage if the print output changes and requires that `losetup` is installed on the system.  Also using things like `popen` can expose details we don't want to other tools and results in a full fork and exec.

My suggestion is you just iterate over the files in `/sys/block/loopX/loop/backing_file`, which is what this tool does anyway. ie.
```
int dirFd = open("/sys/block", O_CLOEXEC | O_DIRECTORY);
if (dirFd < 0)
{
   AI_LOG_SYS_ERROR(errno, "failed to open directory @ '/sys/block'");
   return;
}

DIR *sysBlockDir = fdopendir(dirFd);

struct dirent *entry = nullptr;
while ((entry = readdir(sysBlockDir)) != nullptr)
{
   if (strncmp(d->d_name, "loop", 4) != 0)
      continue;

   std::string name(d->d_name);
   name += "/loop/backing_file";

   int fd = openat(dirFd, name.c_str(), O_RDONLY | O_CLOEXEC);
   if (fd < 0)
   {
      AI_LOG_SYS_ERROR(errno, "failed to open file @ '/sys/block/%s'", name.c_str());
   }
   else
   {
      readBackingStoreFile(fd);
      close(fd);
   }
}

closedir(sysBlockDir);
```

rdkPlugins/Storage/source/Storage.h
```diff
@@ -78,12 +80,123 @@ class Storage : public RdkPluginBase
...
     std::vector<std::string> getDependencies() const override;
 
-    std::vector<LoopMountDetails::LoopMount> getLoopMounts();
-    std::vector<std::unique_ptr<LoopMountDetails>> getLoopDetails();
+private:
+    std::vector<MountProperties> getLoopMounts();
+    std::vector<MountProperties> getBindLoopMounts();
+    std::vector<std::unique_ptr<MountDetails>> getMountDetails();
 
+    template<typename T>
```
bgray-sky, line 95: My preference would be to have these functions in a cpp file, rather than the header if possible.

rdkPlugins/Storage/source/BindLoopMountDetails.h
```diff
@@ -0,0 +1,96 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+/*
+ * File: BindLoopMountDetails.h
+ *
+ */
+#ifndef BINDLOOPMOUNTDETAILS_H
+#define BindLOOPMOUNTDETAILS_H
```
bgray-sky, line 24: Missing captialisation on `#define BindLOOPMOUNTDETAILS_H` - shoud be `#define BINDLOOPMOUNTDETAILS_H`

rdkPlugins/Storage/source/BindLoopMountDetails.cpp
```diff
@@ -0,0 +1,459 @@
...
+    int ret = stat(mMount.fsImagePath.c_str(), &file_stat);
+
+    if (ret < 0) {  
+        AI_LOG_ERROR("failed to get file stat for file '%s'",
+                     mMount.fsImagePath.c_str());
+        return nullptr;
+    }
+
+    // Create reference count file based on unique inode of data.img file
+    std::string refCountFilePath = std::string("/tmp/") + std::to_string(file_stat.st_ino);
```
bgray-sky, line 426: Rather than just in `/tmp`, this should be in the Dobby workspace directory (there use to be a `IDobbyEnv::workspaceMountPath()` call to get the workspace path).  Or it that doesn't exist anymore then suggest prefixing name with something like `/tmp/dobbyLoopDeviceRef_<inode_number>`

rdkPlugins/Storage/source/BindLoopMountDetails.cpp
```diff
@@ -0,0 +1,459 @@
...
+    RefCountFileLock lock(refCountFile);
+
+    // step 2 - if loop device doesn't exist, try and open the source file and
+    // attach to a space loop device 
+    std::string loopDevice = StorageHelper::getLoopDevice(mMount.fsImagePath);
+    int loopDevFd(0);
+
+    if (loopDevice.empty())
+    {
+        loopDevFd = StorageHelper::attachLoopDevice(mMount.fsImagePath, &loopDevice);
```
bgray-sky, line 117: Should we log an error if the reference count is greater than one and loop device doesn't exist.

rdkPlugins/Storage/source/RefCountFile.cpp
```diff
@@ -0,0 +1,150 @@
...
+// -----------------------------------------------------------------------------
+/**
+ *  @brief Increment the reference count in the file.
+ *
+ */
+int RefCountFile::Increment()
+{
+    AI_LOG_FN_ENTRY();
+
+    int ref = Read();
```
bgray-sky, line 83: Do we need a sanity check here that the file is open.

rdkPlugins/Storage/source/RefCountFile.cpp
```diff
@@ -0,0 +1,150 @@
...
+ *  @brief Read the reference count file.
+ *
+ *  @return reference count
+ */
+int RefCountFile::Read()
+{
+    AI_LOG_FN_ENTRY();
+    int ref(0);
+
+    lseek(mFd, 0, SEEK_SET);
```
bgray-sky, line 128: Add checks on the return value for both seek and read and report errors with `AI_LOG_SYS_ERROR(errno, message...)`

For the `read()` call should also wrap in `TEMP_FAILURE_RETRY` macro, ie.
```
ssize_t rd = TEMP_FAILURE_RETRY(read(mFd, &ref, sizeof(ref)));
...
```
```

File: rdkPlugins/Storage/source/RefCountFile.cpp [hunk 8]
```diff
@@ -0,0 +1,150 @@
...
+/**
+ *  @brief Write the reference count to file.
+ *
+ *  @return reference count
+ */
+int RefCountFile::Write(int ref)
+{
+    AI_LOG_FN_ENTRY();
+
+    lseek(mFd, 0, SEEK_SET);
```
bgray-sky, line 145: Same as above, check return values and add `TEMP_FAILURE_RETRY` on the write call.

rdkPlugins/Storage/source/StorageHelper.cpp
jignatius: Yes. You're right. Better not to rely on userspace tools such as losetup. Changed the function to:

`std::string StorageHelper::getLoopDevice(const std::string &backingFile)
{
    AI_LOG_FN_ENTRY();
    
    std::string loopDevice;
    int dirFd = open("/sys/block", O_CLOEXEC | O_DIRECTORY);
    if (dirFd < 0)
    {
        AI_LOG_SYS_ERROR(errno, "failed to open directory @ '/sys/block'");
        return loopDevice;
    }

    DIR *sysBlockDir = fdopendir(dirFd);

    // Loop through all block devices starting with 'loop'
    struct dirent *entry = nullptr;
    while ((entry = readdir(sysBlockDir)) != nullptr)
    {
        if (strncmp(entry->d_name, "loop", 4) != 0)
            continue;

        std::string name(entry->d_name);
        name += "/loop/backing_file";

        int fd = openat(dirFd, name.c_str(), O_RDONLY | O_CLOEXEC);
        if (fd < 0)
        {
            AI_LOG_SYS_ERROR(errno, "failed to open file @ '/sys/block/%s'", name.c_str());
        }
        else
        {
            char buffer[512] = {0};
            int len = read(fd, buffer, sizeof(buffer));
            if (len < 0)
            {
                AI_LOG_SYS_ERROR(errno, "failed to read file @ '/sys/block/%s'", name.c_str());
            }
            else
            {
                if (strncmp(backingFile.c_str(), buffer, backingFile.length()) == 0)
                {
                    loopDevice = std::string("/dev/") + entry->d_name;
                    break;
                }
            }
            close(fd);
        }
    }

    closedir(sysBlockDir);

    AI_LOG_FN_EXIT();
    return loopDevice;
}
`

rdkPlugins/Storage/source/RefCountFile.h
```diff
@@ -0,0 +1,89 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+/*
+ * File: RefCountFile.h
+ *
+ */
+#ifndef DEVELOP_REFCOUNTFILE_H
```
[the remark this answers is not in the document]
jignatius: Good spot.

rdkPlugins/Storage/source/BindLoopMountDetails.h
jignatius: Good spot

rdkPlugins/Storage/source/BindLoopMountDetails.cpp
```diff
@@ -0,0 +1,459 @@
...
+        if ((loopDevFd < 0) || (loopDevice.empty()))
+        {
+            AI_LOG_ERROR("failed to attach file to loop device");
+            return false;
+        }
+    }
+    else{
```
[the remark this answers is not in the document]
jignatius: Ok

rdkPlugins/Storage/source/RefCountFile.cpp
```diff
@@ -0,0 +1,150 @@
...
+ *
+ */
+void RefCountFile::Close()
+{
+    AI_LOG_FN_ENTRY();
+
+    if (mFd)
+    {
+        close(mFd);
+        mFd = 0;
```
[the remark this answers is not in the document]
jignatius: Agreed

rdkPlugins/Storage/source/BindLoopMountDetails.cpp
jignatius: Possible. If the app crashed this could happen. Therefore, we need a way to reset the reference count.

jignatius: It shouldn't happen normally, but I don't we should abandon launching the app.

rdkPlugins/Storage/source/RefCountFile.cpp
jignatius: Can add sanity check here.

rdkPlugins/Storage/source/RefCountFile.cpp
jignatius: Ok

rdkPlugins/Storage/source/RefCountFile.cpp
jignatius: Ok

rdkPlugins/Storage/source/Storage.h
jignatius: I created template functions because I want to do the same thing with different unrelated parameter types:
- CreateMountProperties() can be called with `rt_defs_plugins_storage_data_bindloop_element*` and `rt_defs_plugins_storage_data_loopback_element*`;
- CreateMountDetails() can be called to create a LoopMountDetails or BindLoopMountDetails object depending on how you call it (e.g. `auto bindLoopMount = CreateMountDetails<BindLoopMountDetails>(properties);`)

rdkPlugins/Storage/source/BindLoopMountDetails.cpp
jignatius: I don't think I can call `IDobbyEnv::workspaceMountPath()` inside the Storage plugin. So used this format instead: /tmp/dobbyLoopDeviceRef_<inode_number>

TeknoVenus (commented):
Very minor nitpicks, but overall looks good pending verification on actual hardware.

Please can you update the `README.md` file for the Storage plugin to add details about this functionality?

daemon/lib/source/DobbyManager.cpp
```diff
@@ -184,7 +184,7 @@ void DobbyManager::setupSystem()
     // if we have cgroup RT scheduling enabled in the kernel then also set that
     // to defaults
     std::string cgrpCpuPath = mEnvironment->cgroupMountPath(IDobbyEnv::Cgroup::Cpu);
-    if (!cgrpCpuPath.empty())
+    if (!cgrpCpuPath.empty())
```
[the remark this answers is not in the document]
jignatius: Removed whitespace here. Updated the README.md file for Storage plugin.

rdkPlugins/Storage/source/RefCountFile.cpp
```diff
@@ -0,0 +1,203 @@
...
+ *  @brief Close the file descriptor if it's not already closed.
+ *
+ */
+void RefCountFile::Close()
+{
+    AI_LOG_FN_ENTRY();
+
+    if (mFd >= 0)
+    {
+        close(mFd);
```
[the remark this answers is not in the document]
jignatius: Added a check here.

rdkPlugins/Storage/source/RefCountFile.cpp
```diff
@@ -0,0 +1,203 @@
...
+// -----------------------------------------------------------------------------
+/**
+ *  @brief Decrement the reference count in the file.
+ *
+ */
+int RefCountFile::Decrement()
+{
+    AI_LOG_FN_ENTRY();
+
+    int ref = Read();
```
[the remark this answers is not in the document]
jignatius: The `Read()` function already has sanity checks. Plus there's a check on the value of `ref ` isn't -1 (for error).

TeknoVenus (approved):
LGTM.

Verified on xi1/llama - no regressions seen for existing apps.

PR #191, opened 2022-04-07
Thunder plugin should work in open network mode

TeknoVenus:
### Description
Dobby Thunder plugin should not error if container is run in `open` network mode.

Whilst in production containers should use NAT network mode, for debugging it is often useful to use `open` network mode. The Thunder plugin should not attempt to add firewall rules or set THUNDER_ACCESS to the dobby bridge in this scenario

### Test Procedure
Configure a container with `open` network mode and the Thunder plugin. Container should start without error and should be able to access Thunder

### Type of Change


TeknoVenus:
No issues observed on during testing on Llama @Koczek9

Koczek9 (approved):
LGTM

PR #204, opened 2022-07-15
RDKTV-17814: Support for automatically setting UID/GID of mounts

willtallentiresky:
### Description
Added support for configuration of ownership of mounts to reduce the reliance on external scripts.
The user and group can be changed on file and directory mounts, with the latter able to be applied recursively.

### Test Procedure
Add "mountOwner" definitions in the container config.json as decribed in the README for the storage plugin.

### Type of Change

### DIFF
--- a/rdkPlugins/Storage/README.md
+++ b/rdkPlugins/Storage/README.md
@@ -24,6 +24,9 @@ space "destination".
     }
 }
 ```
+The Storage plugin will only create one loop device for a given source file. If multiple containers share
+the same source file, then the Storage plugin will bind mount the same loop device into the different containers.
+Thus, they share the same private storage.
 
 ### Dynamic Mounts
 Add the following section to your OCI runtime configuration `config.json` file to create dynamic mount. 
@@ -51,8 +54,29 @@ It will mount "source" into container "destination" only if the source exists on
     }
 }
 ```
-The Storage plugin will only create one loop device for a given source file. If multiple containers share
-the same source file, then the Storage plugin will bind mount the same loop device into the different containers. Thus, they share the same private storage.
+
+### Mount Owners
+Add the following section to your OCI runtime configuration `config.json` file to configure mount ownership.
+
+```json
+{
+    "rdkPlugins": {
+        "storage": {
+            "required": true,
+            "data": {
+                "": [
+                    {
+                        "source": "/tmp/test",
+                        "user": "root",
+                        "group": "root",
+                        "recursive": false
+                    }
+                ]
+            }
+        }
+    }
+}
+```
 
 If you already have other RDK plugins in the bundle, then just add the storage plugin. Do not create multiple `rdkPlugin` sections.
 
@@ -112,3 +136,26 @@ For every dynamic mount point the Storage plugin should create, there should be 
 }
 ```
 
+### Creating mount owners
+For every mount owner point the Storage plugin should create, there should be one item in the array of "mountOwner". The options inside this object goes as follows:
+
+| Option              | Value                                                                                                                                   |
+| ------------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| `source`            | Directory or file which is the source of the mount, i.e. on host                                                                        |
+|---------------------| ----------------Below this point there are optionals things, with default value in square brackets "[]"---------------------------------|
+| `user`              | Name of user to change to  (if ommitted, use container default user)                                                                    |
+| `group`             | Name of group to change to  (if ommitted, use container default group)                                                                  |
+| `recursive`         | Whether to change ownership recursively when source is a directory (if ommitted, treated as false, e.g. non-recursive)                  |
+
+#### Example
+```json
+"data": {
+    "mountOwner": [
+        {
+            "source": "/tmp/test"
+        }
+    ]
+}
+```
+
+This will change ownership to the container default user and group, which will be applied non-recusively.
--- a/rdkPlugins/Storage/source/MountOwnerDetails.cpp
+++ b/rdkPlugins/Storage/source/MountOwnerDetails.cpp
@@ -1,0 +1,241 @@
+/*
+* If not stated otherwise in this file or this component's LICENSE file the
+* following copyright and licenses apply:
+*
+* Copyright 2022 Sky UK
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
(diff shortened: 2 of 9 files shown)


rdkPlugins/Storage/README.md
```diff
@@ -51,8 +54,29 @@ It will mount "source" into container "destination" only if the source exists on
...
+### Mount Owners
+Add the following section to your OCI runtime configuration `config.json` file to configure mount ownership.
+
+```json
+{
+    "rdkPlugins": {
+        "storage": {
+            "required": true,
+            "data": {
+                "": [
```
Koczek9, line 67: You lost `mountOwner` here

rdkPlugins/Storage/README.md
```diff
@@ -112,3 +136,26 @@ For every dynamic mount point the Storage plugin should create, there should be
 }
 ```
 
+### Creating mount owners
+For every mount owner point the Storage plugin should create, there should be one item in the array of "mountOwner". The options inside this object goes as follows:
```
Koczek9, line 140: For `mountOwner` Storage plugin doesn't create anything. It just changes ownership. Please say that this WILL change ownership in the host, and that path are resolved to the host. Plus it would be nice to mention that similar functionality is done in Podman https://github.com/containers/podman/commit/874f2327e6ca963edda7cc46819d51048d3d19a8 so it is no total heresy to do such thing ;)

rdkPlugins/Storage/source/MountOwnerDetails.cpp
```diff
@@ -0,0 +1,241 @@
...
+                             userId,
+                             groupId);
+                return false;
+            }
+        }
+        else
+        {
+            changeOwner(mMountOwnerProperties.source, userId, groupId);
+        }
+    }
```
Koczek9, line 178: There should be else here, and return false. If there was some non-existent user selected we cannot set it the message was handled in getOwnerIds so you don't need to give another one but you should fail here as you cannot possibly set wrong user.

rdkPlugins/Storage/source/MountOwnerDetails.cpp
```diff
@@ -0,0 +1,241 @@
...
+    {
+        if (mMountOwnerProperties.recursive)
+        {
+            if (!changeOwnerRecursive(mMountOwnerProperties.source, userId, groupId))
+            {
+                AI_LOG_ERROR("Failed to change owner recursively of '%s' to '%d:%d",
+                             mMountOwnerProperties.source.c_str(),
+                             userId,
+                             groupId);
+                return false;
```
Koczek9, line 171: Use status and return it on the end so we get `AI_LOG_FN_EXIT` log message.

rdkPlugins/Storage/source/MountOwnerDetails.cpp
```diff
@@ -0,0 +1,241 @@
...
+            }
+        }
+        else
+        {
+            changeOwner(mMountOwnerProperties.source, userId, groupId);
```
Koczek9, line 176: Report the status of changeOwner (what if it failed?).

rdkPlugins/Storage/source/MountOwnerDetails.cpp
```diff
@@ -0,0 +1,241 @@
...
+ *  @param[out]  userId  ID corresponding to the configured user name
+ *  @param[out]  groupId ID corresponding to the configured group name
+ *  @return true on success, false on failure
+ */
+bool MountOwnerDetails::changeOwnerRecursive(const std::string& path, uid_t userId, gid_t groupId) const
+{
+    bool success = true;
+    DIR* dir = opendir(path.c_str());
+    if (!dir) {
+        return success;
```
Koczek9, line 198: You are setting success initially to true, then if directory doesn't exist you are succeeding? Is this intentional? Shouldn't we return false as we didn't change the ownership (as it doesn't exist). If this is intentional please add this info into the `@return` field i.e. `true on success or non-existing directory, false on failure` But still I think that failing it would make more sense.

rdkPlugins/Storage/source/Storage.cpp
```diff
@@ -503,6 +493,129 @@ std::vector<DynamicMountProperties> Storage::getDynamicMounts() const
...
+ *
+ *  @param[in]  id          Id we want to map
+ *  @param[in]  mapping     Mapping that should be used
+ *  @param[in]  mapping_len Length of mapping
+ *
+ *  @return if found mapped id, if not found initial id
+ */
+void Storage::setupOwnerIds(uid_t& uid, gid_t& gid) const
+{
+    // Get uid/gid from process
```
Koczek9, line 595: You should pre-set zero here, I know that you did every time you used this method, but in future you cannot guarantee that it will be pre-set by everyone using that method, but its being 0 is required if `mContainerConfig->process` or `mContainerConfig->process->user` are 0.

rdkPlugins/Storage/source/Storage.cpp
```diff
@@ -503,6 +493,129 @@ std::vector<DynamicMountProperties> Storage::getDynamicMounts() const
...
+        AI_LOG_ERROR("No storage data in config file");
+    }
+
+    AI_LOG_FN_EXIT();
+    return mountOwners;
+}
+
+// -----------------------------------------------------------------------------
+/**
+ *  @brief Gets userId and groupId
```
Koczek9, line 585: Can you change the description into more verbose one i.e. `Gets userId and groupId of container owner and maps it into host user namespace.`

Koczek9 (changes requested):

rdkPlugins/Storage/source/MountOwnerDetails.cpp
willtallentiresky: Already reported in changeOwner itself.

rdkPlugins/Storage/source/MountOwnerDetails.cpp
willtallentiresky: Addressed by introducing success flag defaulted to false.

rdkPlugins/Storage/source/MountOwnerDetails.cpp
willtallentiresky: It defaults to true for the purpose of the &= usage.
However, I missed that false should be returned if path is not a directory.
Fixed.

rdkPlugins/Storage/source/Storage.cpp
willtallentiresky: Changed commet to be more accurate but this function does not perform any mapping itself or changing of ownership itself.

Koczek9 (approved):
LGTM

PR #211, opened 2022-10-07
LLAMA-8016 Fixed race with veth devices

Koczek9:
### Description
Fixed the race when one container was closed and second was started at the same time. It resulted with situation when container A still held information about some vethX, while container B was already assigning it to itself, so when container A postHaltHook was done it removed all traces of vethX which made container B loose the network.

### Test Procedure
1. Start any container
2. Copy its network setup to some temp file i.e.
 `cp /tmp/dobby/plugin/networking/sleepy /tmp/dobby/plugin/networking/tmp_name`
3. Close original container
4. Start original container again
5. There should be a message:
` Tried to use already taken vethName 'veth0', continue looking`
and container should take next available veth instead

### Type of Change

### DIFF
--- a/pluginLauncher/lib/include/DobbyRdkPluginUtils.h
+++ b/pluginLauncher/lib/include/DobbyRdkPluginUtils.h
@@ -40,6 +40,7 @@
 #include <list>
 #include <mutex>
 #include <arpa/inet.h>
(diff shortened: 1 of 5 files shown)


Koczek9:
Wrong email address

PR #215, opened 2022-10-24
RDKTV-17862 Fixed order of ipv6 rules

Koczek9:
### Description
Previously the ICMPv6 rule was too low, making it not activated. Now the order is proper which means that communication should work fine.

Before the fix only the unicast address inside container was `REACHABLE`, as the router advertisement went from it, while all other ICMPv6 rules were dropped by the anti-spoofing rule due to them having "improper" address, By changing the the order of operations now we save all neighbor discovery frames.

### Test Procedure
1. Start container with ipv6 networking on
2. Get bash inside container `crun --root /run/rdk/crun exec --tty <container_name> /bin/bash`
3. Check IP address inside container `ifconfig`
4. From outside container try to access local type IPv6 address (starts with `fe80::`) i.e. `ping -c 1 fe80::dccf:8aff:fee1:923c`
5. From outside container check neighbor table `ip -6 neigh` container should be in `REACHABLE` state
6. From inside container check neighbor table `ip -6 neigh` host should be in `REACHABLE` state

### Type of Change

### DIFF
--- a/daemon/lib/source/DobbyManager.cpp
+++ b/daemon/lib/source/DobbyManager.cpp
@@ -505,6 +505,12 @@ void DobbyManager::cleanupContainersShutdown()
             // notified of the container stop event
             if (!stopContainer(it->second->descriptor, false))
             {
+                // As DobbyRunC::killCont already handles problem of masked SIGTERM in
+                // case we failed to stop it means that it tried to SIGKILL too, so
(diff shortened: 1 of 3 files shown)


Koczek9:
Some improper merges clattering the view, will re-upload this commit.

PR #219, opened 2022-11-10
RDKTV-20457 EnvironmentVariable Plugin added

Koczek9:
### Description
We need to be able to add environment variable from host into container. This plugin provides said functionality.

### Test Procedure

1.  Add Environment Variable into the container config i.e.
```
        "environmentvariable": {
            "required": false,
            "data": {
                "variables": [
                    "LANG",
                    "HOME"
                ]
            }
        }
```
2. Start container i.e. `DobbyTool start sleepy ./sleepy_bundle`
3. Open sh inside container `crun --root /run/rdk/crun exec --tty <container_name> /bin/bash`
4. Check if environment variables are set `env`

### Type of Change

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,6 +297,7 @@ option(PLUGIN_HTTPPROXY "Include HttpProxy plugin" OFF)
 option(PLUGIN_APPSERVICES "Include AppServices plugin" OFF)
 option(PLUGIN_IONMEMORY "Include ION Memory plugin" OFF)
 option(PLUGIN_DEVICEMAPPER "Include DeviceMapper plugin" OFF)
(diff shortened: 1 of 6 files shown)


Koczek9:
The issue is due to the fact that link for  "/etc/localtime -> /usr/share/zoneinfo/Universal" is incorrect. The "TZ" environment variable is set only for the WPEWebProcess and getting it from there would be extremely hacky way of fixing. We should make "/etc/localtime" correct inside "LocationSync" rdkservice for whole box (probably by using "tzest" after setting envvar) and after that we can just use "LocalTime" Dobby plugin.
Also in here there would be problems with environment variables that would get added after DobbyDaemon is started - they would not be available to DD process. As DD is started very early in booting procedure whole plugin idea is faulted. So for now closing this pull request.

PR #221, opened 2022-11-22
RDKTV-20457: Added setTZ param to localTime plugin

mateusz-hobgarski-red:
### Description
Added setTZ param to localTime plugin

### Test Procedure
Set setTZ param in localTime plugin of sleepy container, check logs if Dobby reads it

### Type of Change

### DIFF
--- a/rdkPlugins/LocalTime/source/LocalTimePlugin.cpp
+++ b/rdkPlugins/LocalTime/source/LocalTimePlugin.cpp
@@ -32,7 +32,8 @@ LocalTimePlugin::LocalTimePlugin(std::shared_ptr<rt_dobby_schema> &containerConf
                                  const std::string &rootfsPath)
     : mName("LocalTime"),
       mRootfsPath(rootfsPath),
-      mContainerConfig(containerConfig)
+      mContainerConfig(containerConfig),
+      mUtils(utils)
 {
     AI_LOG_FN_ENTRY();
     AI_LOG_FN_EXIT();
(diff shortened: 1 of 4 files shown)


rdkPlugins/LocalTime/source/LocalTimePlugin.cpp
```diff
@@ -83,6 +85,54 @@ bool LocalTimePlugin::postInstallation()
     return true;
 }
 
+// -----------------------------------------------------------------------------
+/**
+ *  @brief postInstallation OCI hook.
+ *
+ *  If set_tz parameter is set then its value should be a path to file.
```
Koczek9, line 92: It is interesting that in schema file you are naming it `setTZ` and here `set_tz` and it works. But it seams that files from schema doesn't like capital letters :). What matters is that it works.

rdkPlugins/LocalTime/source/LocalTimePlugin.cpp
```diff
@@ -83,6 +85,54 @@ bool LocalTimePlugin::postInstallation()
...
+        {
+            std::string tz;
+            std::getline(file, tz);
+            mUtils->addEnvironmentVar("TZ=" + tz);
+
+            AI_LOG_DEBUG("read from set_tz: %s", tz.c_str());
+        }
+        else
+        {
+            AI_LOG_WARN("unable to open '%s'", set_tz);
```
Koczek9, line 123: This should be `AI_LOG_ERR` as you are failing, but as this is the only wrong thing I would not make a commit for fixing just this one thing

Koczek9 (approved):
All in all LGTM

goruklu (approved):

goruklu (approved):

PR #229, opened 2023-03-20
ONEM-29861 Some DAC apps do not start: fix container state read

sverkoye:
Status was read into too small buffer and so parsing failed. Made bigger buffer and added logging to indicate better error for future "too small" buffer.

### Description
Problem happened with container that has several extra annotations. They seem to be returned with reading container state and the while state was > 1000 bytes.

### Test Procedure
Fix can be verified with a container with extra annotations

### Type of Change
- [X] Bug fix (non-breaking change which fixes an issue)

- only SRCREV


goruklu (approved):

goruklu:
thanks @sverkoye

PR #282, opened 2023-11-08
RDK-45298: Checking /opt/dobby.json for settings

mateusz-hobgarski-red:
### Description
Added checking /opt/dobby.json for dobby settings.

### Test Procedure
Add /opt/dobby.json config file. Check in Dobby logs if it was open instead of /etc/dobby.json.

### Type of Change


daemon/process/source/Main.cpp
```diff
@@ -253,8 +253,21 @@ static void parseArgs(int argc, char **argv)
  */
 static std::shared_ptr<Settings> createSettings()
 {
-    // create settings from either defaults or JSON settings file
+    // create settings from either defaults, development or JSON settings file
     std::shared_ptr<Settings> settings;
+
+#if (AI_BUILD_TYPE == AI_DEBUG)
```
goruklu, line 259: currently we can't rely on AI_BUILD_TYPE=AI_DEBUG to distinguish between PROD and VBN builds. We need a dedicated compile flag to allow "/opt/dobby.json"

goruklu (changes requested):

daemon/process/source/Main.cpp
mateusz-hobgarski-red: Added ENABLE_OPT_SETTINGS compilation flag.

PR #284, opened 2023-11-16
RDK-45396 Avoid to run the L1 tests in every build job

sithanda:
### Description
1. Run all the tests in the github workflow by adding a new flag (RUN_TESTS => which will enable all -DLEGACY_COMPONENTS=ON -DRDK=ON -DUSE_SYSTEMD=ON these profiles in debug build type) to the yml file. 
Therefore, two additional builds have been included to the compilation that enable all profiles and It will only build if the release build type is selected and the RUN_TESTS is not added.
 
2. Updated the cmakelist.txt file with including only the required files are compiled and the not the whole dobby source code

### Test Procedure
In github workflow L1 test is running and passing

### Type of Change
- Enhancing Dobby L1 test

- Not required


.github/workflows/build.yml
```diff
@@ -58,17 +59,24 @@ jobs:
          cd $GITHUB_WORKSPACE
          mkdir build
          cd build
-         cmake -DCMAKE_TOOLCHAIN_FILE="${{ env.TOOLCHAIN_FILE }}" -DRDK_PLATFORM=DEV_VM -DCMAKE_INSTALL_PREFIX:PATH=/usr -DENABLE_DOBBYL1TEST=ON -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} ${{ matrix.extra_flags }} ${{ env.optional_plugins }} ..
+         if [ $matrix.extra_flags = "RUN_TESTS" ]
+         then
+             cmake -DCMAKE_TOOLCHAIN_FILE="${{ env.TOOLCHAIN_FILE }}" -DRDK_PLATFORM=DEV_VM -DCMAKE_INSTALL_PREFIX:PATH=/usr -DENABLE_DOBBYL1TEST=ON -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} ${{ matrix.optional_flags }} ${{ env.optional_plugins }} ..
+         else
+             cmake -DCMAKE_TOOLCHAIN_FILE="${{ env.TOOLCHAIN_FILE }}" -DRDK_PLATFORM=DEV_VM -DCMAKE_INSTALL_PREFIX:PATH=/usr -DENABLE_DOBBYL1TEST=ON -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} ${{ matrix.extra_flags }} ${{ env.optional_plugins }} ..
```
goruklu, line 66: Why do we need to enable L1 test build option (ENABLE_DOBBYL1TEST)  if we are not going to run it ?

goruklu (changes requested):

.github/workflows/build.yml
sithanda: Yes, Not required to enable ENABLE_DOBBYL1TEST option if we are not running the test. Now I have removed it. thanks

goruklu (approved):

PR #311, opened 2024-01-25
RDK-46375-Include the L2 Test in Github Action

sithanda:
### Description
Changes Done:
1. Changed L2 test folder architecture, moved L2 test cases from tests to tests/L2_testing
2. skipped gui_containers when run in vm
3. Added test passing count in logs
4. Enabled only passing tests in runner.py
            -  basic_sanity_tests.py
            -  container_manipulations.py
            -  bundle_generation.py
            -  plugin_launcher.py
6. Fixed SyntaxWarning: invalid escape sequence

### Test Procedure
1. Run in vagrant setup
2. Run in github workflow.

### Type of Change
- [ ] L2 test enhancement

- [ ] NA


unit_tests/L2_testing/test_runner/runner.py
```diff
@@ -25,40 +25,43 @@
...
                    container_manipulations,
-                   command_line_containers,
-                   start_from_bundle,
                    bundle_generation,
                    plugin_launcher,
-                   thunder_plugin,
-                   gui_containers,
-                   network_tests]
-
+                   gui_containers]
```
goruklu, line 36: why are you removing command_line_containers, start_from_bundle and thunder_plugin tests and keeping gui_containers ? We need to do the opposite, no ?

goruklu (changes requested):

goruklu:
also, can we rename the "unit_tests" directory to "testing" or just "tests" and then have L1_testing and L2_testing directories under that ?
Having L2_testing directory under "unit_tests" sound a bit wrong.

unit_tests/L2_testing/test_runner/runner.py
sithanda: The tests for thunder plugin, start from bundle, network tests, command line containers, and start from bundle are not passing, and we are trying to resolve this. The github workflow will show failiure if those are added to runner.py, we will add those tests after the failures have been fixed.

The gui_containers test will not launch when it is executed in a virtual platform(vagrant/github workflow). so need to remove from here.

goruklu (approved):

PR #333, opened 2024-08-29
[RDK-52502] added new mount/unmount APIs

goruklu:
initial implementation of mount & unmount operations

### Description
What does this PR change/fix and why?
Adds new mount & unmount API calls to support dynamic mounting/unmounting of USB mass storage devices

### Test Procedure
Use the following commands to mount and unmount directories in already running containers:
```
DobbyTool mount com.sky.as.apps_com.bskyb.epgui  /mnt/usb /tmp/newmount ro,bind
DobbyTool unmount com.sky.as.apps_com.bskyb.epgui /tmp/newmount
```

### Type of Change

- [x] The base Bitbake recipe (`meta-rdk-ext/recipes-containers/dobby/dobby.bb`) must be modified to support the changes in this PR (beyond updating `SRC_REV`)

### DIFF
--- a/daemon/lib/source/Dobby.cpp
+++ b/daemon/lib/source/Dobby.cpp
@@ -650,6 +650,8 @@ void Dobby::initIpcMethods()
         {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_RESUME,                  &Dobby::resume                 },
         {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_HIBERNATE,               &Dobby::hibernate              },
         {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_WAKEUP,                  &Dobby::wakeup                 },
+        {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_MOUNT,                   &Dobby::addMount                  },
+        {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_UNMOUNT,                 &Dobby::removeMount                },
         {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_EXEC,                    &Dobby::exec                   },
         {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_GETSTATE,                &Dobby::getState               },
         {   DOBBY_CTRL_INTERFACE,        DOBBY_CTRL_METHOD_GETINFO,                 &Dobby::getInfo                },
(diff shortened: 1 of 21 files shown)


goruklu:
requires kernels 4.18 or higher since this implementation requires SYS_open_tree and SYS_move_mount syscalls.

goruklu:
need to add L1/L2 tests

daemon/lib/source/Dobby.cpp
```diff
@@ -1436,6 +1438,118 @@ void Dobby::wakeup(std::shared_ptr<AI_IPC::IAsyncReplySender> replySender)
...
+ *
+ *
+ *
+ *
+ */
+void Dobby::addMount(std::shared_ptr<AI_IPC::IAsyncReplySender> replySender)
+{
+    AI_LOG_FN_ENTRY();
+
+    // Expecting 4 arguments  (int32_t cd, string source, string target, int32_t mountFlags)
```
bgray-sky, line 1453: I think the API should be `addMount(int32_t cd, string source, string target, vector<string> mountFlags, string mountData)`

Most notably the mount flags is a string list taking the same options as the mount command line app.  That's kinda what we've done for the spec file and potentially avoids different values for the mount flags.

daemon/lib/source/DobbyManager.cpp
```diff
@@ -1735,6 +1738,287 @@ bool DobbyManager::wakeupContainer(int32_t cd)
...
+    if (containerPid == 0)
+    {
+        AI_LOG_WARN("failed to find container pid for container with descriptor %d", cd);
+        AI_LOG_FN_EXIT();
+        return false;
+    }
+    AI_LOG_INFO("containerPid: %d", containerPid);
+ 
+    // fork the process
+    pid_t pid = fork();
```
bgray-sky, line 1785: Using fork is not ideal, presumably this was done because of the userns switch and setting the UID / GID for the `mkdir` command.  So I did an experiment and don't think it's necessary to switch into the userns which means I think you can implement this simply via the `IDobbyUtils::callInNamespace` API.

The trick is you need to switch into the uid / gid of the container, and you can do that with the `SYS_setresgid` and `SYS_setresuid` which set the effective UID / GID on a single thread.  So should be able to do the following:
```

int mountFd = open_tree(...

auto lambda = [mountFd, containerUid, containerGid]()
{
    // switch to uid / gid of the container
    if (syscall(SYS_setresgid, -1, containerUid, -1) != 0)
       AI_LOG_ERROR(....
    if (syscall(SYS_setresuid, -1, containerGid, -1) != 0)
        AI_LOG_ERROR(....

    mkdir(mountPoint ...);

    // switch back to root for the mount
    if (syscall(SYS_setresgid, -1, 0, -1) != 0)
        AI_LOG_ERROR(....
    if (syscall(SYS_setresuid, -1, 0, -1) != 0)
        AI_LOG_ERROR(....

    move_mount(fdMount, ...)
};

mUtilities->callInNamespace(pid, CLONE_NEWNS, lambda);

close(fdMount);
```

daemon/lib/source/DobbyManager.cpp
```diff
@@ -1735,6 +1738,287 @@ bool DobbyManager::wakeupContainer(int32_t cd)
...
+    {
+        AI_LOG_WARN("failed to find container pid for container with descriptor %d", cd);
+        AI_LOG_FN_EXIT();
+        return false;
+    }
+    
+    AI_LOG_INFO("containerPid: %d", containerPid);
+
+    // fork the process
+    pid_t pid = fork();
```
bgray-sky, line 1940: Same as above, I think you can avoid the `fork` call by switching the effective uid / gid on the thread running in the mount namespace.

daemon/lib/source/DobbyManager.cpp
goruklu: @bgray-sky yes, I had to use fork() because of userns. I'll try your suggestion  which looks like a much cleaner way, thanks!

goruklu: I've removed the fork() and replaced it with setresgid/uid and  callinnamespace() calls. It works!

daemon/lib/source/DobbyManager.cpp
goruklu: I've removed the fork() and replaced it with setresgid/uid and  callinnamespace() calls. It works!

daemon/lib/source/Dobby.cpp
goruklu: I was planning to use  syscall(SYS_mount_setattr,...) for the mount flags/attirbutes but it's not supported in kernel v5.4 Kirkstone devices have.

goruklu: using     MS_REMOUNT to set the mount flags instead of SYS_mount_setattr works.

goruklu: addMount() is now taking a string for the mountFlags argument. it's a comma seperated list of mount options (ro,bind etc).

goruklu:
replaced with #336

PR #343, opened 2024-10-02
[RDK-53519] added mount tunnel based addMount() implementation

goruklu:
### Description
add support for dynamic mounting of devices/dirs inside containers on devices with older linux kernels (<5.4)

### Test Procedure
Container config should have `rootfsPropagation` set to `slave` mode.

Use the following commands to mount and unmount directories in already running containers:

```
DobbyTool mount com.sky.as.apps_com.bskyb.epgui  /mnt/usb /tmp/newmount ro,bind
DobbyTool unmount com.sky.as.apps_com.bskyb.epgui /tmp/newmount
```

### Type of Change


daemon/lib/source/DobbyManager.cpp
```diff
@@ -1924,9 +1924,110 @@ bool DobbyManager::addMount(int32_t cd, const std::string &source, const std::st
...
+            return false;
+        }
+
+        if (syscall(SYS_setresuid, -1, 0, -1) != 0)
+        {
+            AI_LOG_ERROR("failed to setresuid for root");
+            return false;
+        }
+        // move the mount from the temporary mount point inside the container to the final mount point
+        if(mount(tempMountPointInsideContainer.c_str(), mountPointInsideContainer.c_str(), nullptr, mountOptions | MS_MOVE, nullptr))
```
bgray-sky, line 1982: I don't think you need `mountOptions | MS_MOVE` here, just `MS_MOVE` will do, as according to the mount man page:

>  Moving a mount
> ...
> The remaining bits in the mountflags argument are ignored, as are the filesystemtype and data arguments.

daemon/lib/source/DobbyManager.cpp
```diff
@@ -1924,9 +1924,110 @@ bool DobbyManager::addMount(int32_t cd, const std::string &source, const std::st
...
+    // this is needed to move the mount inside the container namespace later
+    if(mount(source.c_str(), tempMountPointOutsideContainer.c_str(), nullptr, mountOptions, nullptr))
+    {
+        AI_LOG_WARN("mount failed for %s errno: %d", destination.c_str(), errno);
+        mUtilities->rmdirRecursive(tempMountPointOutsideContainer.c_str());
+        AI_LOG_FN_EXIT();
+        return false;
+    }
+
+    auto doMoveMountLambda = [containerUID, containerGID, tempMountPointInsideContainer, mountPointInsideContainer, mountOptions, mountData]()
```
bgray-sky, line 1949: Don't think you need to capture `mountOptions` and `mountData` in the lambda, not that it hurts.

daemon/lib/source/DobbyManager.cpp
```diff
@@ -1924,9 +1924,110 @@ bool DobbyManager::addMount(int32_t cd, const std::string &source, const std::st
...
+     if (!mUtilities->mkdirRecursive(tempMountPointOutsideContainer.c_str(), 0755))
+     {
+            AI_LOG_ERROR("failed to create temporary mount point %s", tempMountPointOutsideContainer.c_str());
+            AI_LOG_FN_EXIT();
+            return false;
+     }
+    
+    // mount the source dir on the temporary mount point outside the container
+    // this is needed to move the mount inside the container namespace later
+    if(mount(source.c_str(), tempMountPointOutsideContainer.c_str(), nullptr, mountOptions, nullptr))
```
bgray-sky, line 1941: Can we add `mountData` as the last parameter to `mount(...)`.

rdkPlugins/Storage/source/Storage.cpp
```diff
@@ -89,6 +94,28 @@ bool Storage::preCreation()
...
+        return false;
+    }
+    // Create directory inside the container rootfs for the mount tunnel
+    if (!DobbyRdkPluginUtils::mkdirRecursive(mMountPointInsideContainer, 0755))
+    {
+        AI_LOG_WARN("failed to create dir '%s'", mMountPointInsideContainer.c_str());
+        return false;
+    }
+
+    if(mount(mTempMountPointOutsideContainer.c_str(), mTempMountPointOutsideContainer.c_str(), NULL, MS_BIND, NULL) != 0)
```
bgray-sky, line 112: Just checking - is this line right ?  It's bind mounting `mTempMountPointOutsideContainer` back to `mTempMountPointOutsideContainer` ?

daemon/lib/source/DobbyManager.cpp
goruklu: removed mountOptions, thanks

daemon/lib/source/DobbyManager.cpp
goruklu: added mountData to the mount() call here.

rdkPlugins/Storage/source/Storage.cpp
goruklu: yes we need the bind mount here, otherwise mount propagation does not work.  
From: https://brauner.io/2023/02/28/mounting-into-mount-namespaces.html
```
mount --bind /opt /opt
mount --make-private /opt
mount --make-shared /opt
# Create new mount namespace with all mounts turned into dependent mounts.
unshare --mount --propagation=slave
```

CMakeLists.txt
```diff
@@ -168,6 +168,7 @@ find_package( breakpad QUIET )
 find_path(LINUXMOUNT NAMES "linux/mount.h")
 if(NOT LINUXMOUNT)
```
bgray-sky, line 169: I don't think the test for `linux/mount.h` is sufficient to test if `open_tree` syscall is supported.  It may have to be done at runtime using `uname` or maybe as an explicit build flag ?

FWIW - I found the `linux/mount.h` was available in toolchains built for platforms running 4.9 kernels, and if built and run on one of those platforms I get the following error when trying to add a mount:
```
ERR: open_tree failed errno: 38 container: 734
```

ks734 (approved):
LGTM

PR #366, opened 2025-02-11
RDKTV-35344: Dynamic mounts are failing to unmount

ks734:
### Description
Removing the dynamic mounts in postStop hook, since it is failing to unmount when the container is stopping

### Type of Change

### DIFF
--- a/rdkPlugins/Storage/source/DynamicMountDetails.cpp
+++ b/rdkPlugins/Storage/source/DynamicMountDetails.cpp
@@ -211,7 +211,7 @@ bool DynamicMountDetails::onPostStop() const
 
     if (stat(targetPath.c_str(), &buffer) == 0)
     {
-        if (umount(targetPath.c_str()) == 0)
+        if (remove(targetPath.c_str()) == 0)
         {
             success = true;
         }


goruklu (commented):
@ks734 why do we create the target dir/file both in onCreateRuntime() and onCreateContainer() hooks ? Am I missing something ?

ks734:
> @ks734 why do we create the target dir/file both in onCreateRuntime() and onCreateContainer() hooks ? Am I missing something ?

In onCreateContainer() hook, we are checking whether the target destination path already exists, https://github.com/rdkcentral/Dobby/blob/master/rdkPlugins/Storage/source/DynamicMountDetails.cpp#L138 
Only if the target dir/file does not exist, we create it and then the mount is added.

goruklu:
I still don't understand why we need to check the target dir and create it if it's not present in createContainer(). Is there a case where target dir is not created in createRuntime() ?

ks734:
> I still don't understand why we need to check the target dir and create it if it's not present in createContainer(). Is there a case where target dir is not created in createRuntime() ?

Yes, for /tmp directory, it is not created in createRuntime() hook and it is failing in createContainer() hook when we are trying to addMount()

goruklu (approved):

PR #384, opened 2025-07-30
RDKEMW-7515: Add custom userid,groupid support to bundle path

madanagopalt:
### Description
This change adds support to pass custom userid,groupid to container in bundle path instead of root always. 
 
### Test Procedure
Ensure container works with bundle path

### Type of Change

- No


goruklu (changes requested):
@madanagopalt can you please fix the Coverity findings and the failing L1tests ?
also, can you confirm these changes are backwards compatible and will not break existing builds (calling startfrombundle() without UID/GID parameters)?

goruklu (dismissed):

